PR #2, opened 2020-07-28
Code reorganisation

ialarmedalien:
Further reorganisation of the RE api and spec directories to make the code a little more integrated.

Common files (e.g. CODEOWNERS, .travis.yaml, Makefile, Dockerfiles) moved to the top level and merged where appropriate.

Basic structure of the docker images created by the Dockerfiles:

```
app
+-- client_src
+-- importers
+-- relation_engine_server
+-- scripts   => run_test, start_server, etc.
+-- spec      => json schema and tests
```
All the dirs now have their tests within the dir.


jayrbolton:
I only have one requested change:
For tests, we have `/test` and also `src/test` that we could consolidate into something like `test/api` and `test/spec`

Side comment:
While this is becoming more of a monorepo, I don't want it to become a monolith. One key difference is that we can still have discrete packages with separate concerns and dependencies. For example, we want to be able to run the client with a separate set of dependencies from the server. In the future, we should use Poetry for dependency management for the python code, and we can have a pyproject.toml file in src/server/pyproject.toml, src/client/pyproject.toml, and src/importers/pyproject.toml

Also, we should namespace API stuff a little bit more and not assume it owns some kind of global namespace. For example, scripts/local-build.sh is for the API, but if we ever want a similar script for another service, such as a kafka listener, then this would get a little messy and confusing.

ialarmedalien:
In light of your comments, I think it might be better to have the repo organised such that each part of the project has its tests within the dir that the code is in, i.e. `relation_engine_server` has a subdirectory `tests`, `spec` has a subdirectory `tests`, etc.

New top level dir structure:
```
app
+-- client_src
+-- importers
+-- relation_engine_server
+-- scripts   => run_test, start_server
+-- spec      => json schema and tests
```

Regarding building docker images, creating poetry packages of different bits of code, creating GitHub exports, etc., if we decide which parts of the repo it makes sense to provide on their own, we can create packages / exports of them. Docker image builds and uploads to Dockerhub can be easily moved over to Github Actions and could be done automatically on merge to the main branch or when creating a new release.

relation_engine_server/utils/pull_spec.py
```diff
@@ -125,7 +122,4 @@ def _save_release_id(info):
 
 
 if __name__ == '__main__':
-    if len(sys.argv) > 1:
-        if sys.argv[1] == 'init':
-            download_specs(init_collections=True)
```
ialarmedalien, line 130: `init_collections` is true by default, so this just runs the same command twice.

spec/test/stored_queries/test_list_test_vertices.py
```diff
@@ -40,11 +30,18 @@ def setUpClass(cls):
...
+            [
+                {'is_public': True, '_key': 'a', 'ws_id': 10},  # public access
+                {'is_public': False, '_key': 'b', 'ws_id': 1},  # private access
+                {'is_public': False, '_key': 'c', 'ws_id': 99}  # no access
+            ]
+        )
+        self.assertEqual(
+            docs_created.json(),
+            {'created': 3, 'details': [], 'empty': 0, 'error': False, 'errors': 0, 'ignored': 0, 'updated': 0},
+        )
```
ialarmedalien, line 44: turning this into a test, rather than just printing some output

CODEOWNERS
```diff
@@ -1,3 +1,3 @@
 * @jayrbolton
 
-api/* @jayrbolton @slebras
```
jayrbolton, line 3: Feel free to add yourself here. Mainly it auto-sets you as reviewer for PRs, and also serves as documentation for other devs

docker-compose.yaml
```diff
@@ -4,16 +4,19 @@ version: '3'
...
     build:
       context: .
+      dockerfile: Dockerfile
       args:
         DEVELOPMENT: 1
     ports:
       - "127.0.0.1:5000:5000"
-    volumes:
-      - ${PWD}:/app
+#    uncomment to mount local directories
```
jayrbolton, line 16: Why is this commented?

scripts/local-build.sh
```diff
@@ -2,6 +2,6 @@
 set -e
 # show the commands we execute
 set -o xtrace
-export IMAGE_NAME="kbase/relation_engine_api:0.0.5"
+export IMAGE_NAME="kbase/relation_engine_api:local_build"
```
jayrbolton, line 5: Here's the approach I've been using lately: https://github.com/kbase/search_api2/blob/develop/scripts/docker_deploy

I set a canonical version in one place in `/VERSION` and refer to that when building and deploying.

jayrbolton:
Regarding this:

```
app
+-- client_src
+-- importers
+-- relation_engine_server
+-- scripts   => run_test, start_server
+-- spec      => json schema and tests
```

I'm not sure what this syntax means... is app a top-level directory that everything is underneath? Do we need a directory called "app"?

Otherwise everything in that comment sounds good to me

This PR looks good to me as a stepping stone for more PRs

docker-compose.yaml
ialarmedalien: I had been using it during development but then commented it out whilst ensuring tests were passing, etc.

CODEOWNERS
ialarmedalien: ok, will do on the next PR.

ialarmedalien:
`app` is a top level directory in the docker image -- we don't have to add it, it gets created in the Dockerfile.

No need to worry, there are more PRs coming up!

scripts/local-build.sh
ialarmedalien: I'll put in an issue to migrate to that syntax in this repo.

PR #6, opened 2020-07-31
Consolidating all JSON validation and "wait for..." actions

ialarmedalien:
- consolidate all JSON validation functions into `relation_engine_server.utils.json_validation`, so the same validator is used throughout the codebase
- move all the "wait_for..." code into `relation_engine_server.utils.wait_for`, instead of having similar functions in different places
- ensure that all the tests have the appropriate services running prior to execution
- create schema files to validate against for the stored queries, views, and collections, instead of having the info hard-coded in a file


relation_engine_server/utils/wait_for.py
```diff
@@ -0,0 +1,89 @@
+"""
+Block until all dependent services come online.
+"""
+import requests
+import time
+import sys
+from .config import get_config
```
jayrbolton, line 7: Let's use the full import paths, so this would be `from relation_engine_server.utils.config import get_config`

relation_engine_server/utils/wait_for.py
```diff
@@ -0,0 +1,89 @@
...
+
+    while True:
+        try:
+            for service in service_conf_list:
+                name = service['name']
+                url = service['url']
+                if service['auth'] is not None:
+                    requests.get(service['url'], auth=service['auth']).raise_for_status()
+                else:
+                    requests.get(service['url'])
```
jayrbolton, line 26: This line doesn't have the `raise_for_status()` call like line 24

spec/collection_schema.yaml
```diff
@@ -0,0 +1,13 @@
+name: collection_schema
+type: object
+required: ['name', 'type', 'schema']
+properties:
+  name:
```
jayrbolton, line 5: There are also these fields:
* delta (boolean) -- whether this collection uses "time travel" features (has "id", "from", "to" fields)
* indexes (array of objects with properties for "type" (string/enum) and "fields" (array of string))

jayrbolton: I would also add `additionalProperties: false` here

spec/stored_query_schema.yaml
```diff
@@ -0,0 +1,12 @@
+name: stored_query_schema
+type: object
+required: ['query', 'name']
```
jayrbolton, line 3: Let's add "description" (string) in here, and also add `additionalProperties: false`

spec/test/helpers.py
```diff
@@ -15,53 +13,22 @@ def get_config():
...
-        except Exception as err:
-            print(f"Waiting for dependent service to come online: {url}")
-            if int(time.time()) > timeout:
-                sys.stderr.write(str(err) + "\n")
-                raise RuntimeError(f"Timed out waiting for {url}")
-            time.sleep(2)
+def run_query(query_name, query_data={}):
+    """submit a database query"""
+
+    query_results_url = os.environ['RE_API_URL'] + '/api/v1/query_results'
```
jayrbolton, line 25: wouldn't you want to use `get_config()['re_query_results_url']` here?

relation_engine_server/utils/wait_for.py
ialarmedalien: That was how it was in the original version -- I don't know it there was a particular reason for that or not. If you aren't aware of there being some reason to not use `raise_for_status`, I'll put it in.

spec/stored_query_schema.yaml
ialarmedalien: There's an issue using `additionalProperties: false` here because the API injects `ws_ids` into the query data (see https://github.com/kbase/relation_engine/blob/develop/relation_engine_server/api_versions/api_v1.py#L53 onwards). I don't like it, as it's a bit too "magical", plus it breaks the JSON schema validation if you try to enforce the `additionalProperties` rule. Either the `ws_ids` should be declared in the (relevant) schema files or json schema validation should occur before they get injected.

jayrbolton: ws_ids is injected into the parameters for the query, but this schema is validating the definition file for the query. It seems like you are talking about having `additionalProperties: false` inside the schema for the params for an individual query, but I'm talking about putting it in the schema for the query definition

ialarmedalien: ah sorry, gotcha.

spec/test/helpers.py
ialarmedalien: yes! Indeed, that is why I created that variable. 🤦

relation_engine_server/utils/wait_for.py
jayrbolton: I don't remember a particular reason... maybe I was relying on it raising a ConnectionError. But it makes sense to me to use an endpoint where we expect a 2xx response and use raise_for_status

ialarmedalien: Turns out that if you do `raise_for_status`, the tests never proceed as the `auth` and `workspace` servers return 500.

spec/test/test_validate.py
```diff
@@ -0,0 +1,191 @@
+"""
```
ialarmedalien, line 1: New file -- testing validation of the different types of specs.

spec/validate.py
```diff
@@ -0,0 +1,199 @@
+"""
```
ialarmedalien, line 1: This was `spec/test/validate.py`; I have moved it up a level (out of `test`) and refactored it.

relation_engine_server/utils/json_validation.py
```diff
@@ -22,7 +25,92 @@ def set_defaults(validator, properties, instance, schema):
...
+                                  defaults to the root of the schema object if not set
+
+    only one of `schema` and `schema_file` should be supplied
+
+    :return:
+    validator  (Validator)  jsonschema validator instance
+
+    """
+
+    if schema is None and schema_file is None or schema is not None and schema_file is not None:
```
jayrbolton, line 53: I'm not asking you to change this, but this could also be `if (schema is None) == (schema_file is None)`, which is a negated xor. Just thought that was kinda interesting :p

relation_engine_server/utils/json_validation.py
```diff
@@ -22,7 +25,92 @@ def set_defaults(validator, properties, instance, schema):
...
+        raise ValueError("Please supply either a schema or a schema file path")
+
+    # schema to validate against
+    if schema is None:
+        schema = _load_json_schema(schema_file)
+
+    # get the appropriate location in the schema
+    validation_schema = resolve_pointer(schema, validate_at)
```
jayrbolton, line 61: Not needed for this PR, but would be useful if `validate_at` could be a path such as "definitions/whatever"

jayrbolton: Disregard

spec/test/helpers.py
```diff
@@ -93,6 +62,15 @@ def create_test_docs(coll_name, docs, update_on_dupe=False):
     return resp
 
 
+def capture_stdout(function, *args, **kwargs):
```
jayrbolton, line 65: Cool!

spec/test/test_validate.py
jayrbolton: Very thorough!!

jayrbolton:
LGTM

PR #8, opened 2020-07-31
Using "collection" to refer to collections, rather than "schema"

ialarmedalien:
What are currently called 'schemas' are the schemas for collections, so rename them.

Change references in the code from `schema` to `collection` wherever appropriate, and rename the `schemas` dir to `collections`.

Added in an `/specs/collections` endpoint to retrieve collection details that mirrors the existing `/specs/schemas` endpoint; ideally at some point, the `/specs/schemas` endpoint can be deprecated in favour of `/specs/collections`.

Also trimmed most of the content from the `spec.tar.gz` file that is used in tests, so that it now only contains schemas that are used in the tests.


jayrbolton:
CHanged base to develop now that "centralise_validation" has been merged into develop

relation_engine_server/README.md
```diff
@@ -264,12 +264,14 @@ _Query params_
 
 Every call to update specs will reset the spec data (do a clean download and overwrite).
 
-### GET /api/v1/specs/schemas
+### GET /api/v1/specs/collections or /api/v1/specs/schemas
```
jayrbolton, line 267: Let's mark `/api/v1/specs/schemas` as deprecated in the docs.

jayrbolton: I think we can remove that path from the above header and just put a deprecation notice in this section somewhere (eg towards the bottom)

relation_engine_server/utils/config.py
```diff
@@ -45,10 +45,10 @@ def get_config():
             'release_id': os.path.join(spec_path, '.release_id'),
             'root': spec_path,
             'repo': spec_repo_path,
-            'schemas': spec_schemas_path,
+            'collections': collections_path,
             'stored_queries': stored_queries_path,
-            'views': spec_views_path,
-            'vertices': os.path.join(spec_schemas_path, 'vertices'),
+            'views': views_path,
+            'vertices': os.path.join(collections_path, 'vertices'),  # unused
```
jayrbolton, line 51: Let's remove this entry ("vertices")

We used to have a directory split of "vertices" and "edges", but later we removed that

relation_engine_server/utils/spec_loader.py
```diff
@@ -69,6 +69,16 @@ def _get_file_name(path):
     return os.path.splitext(os.path.basename(path))[0]
 
 
+class CollectionNonexistent(Exception):
```
jayrbolton, line 72: Seems like you can remove the SchemaNonexistent class now (further below)

jayrbolton:
Que bueno

PR #12, opened 2020-08-11
Extend json validation

ialarmedalien:

Adds tests for URI, date, and string format validation, and moves existing similar tests to `test_json_validation.py`.

Extends the `jsonschema` `RefResolver` class to allow reading of yaml files as well as json.
`RefResolver` now automatically assumes that any `$ref` without an explicit scheme (`http`, `file`, etc.) should be prefixed with `file://`.


relation_engine_server/test/data/json_validation/unquoted_date.yaml
```diff
@@ -0,0 +1,3 @@
+name: unquoted_date
+distance: 3
+date: 2020-06-06
```
ialarmedalien, line 3: Added a test for this, since it caused so much trouble!

relation_engine_server/test/data/schema_refs/level_1/level_2/edge.yaml
```diff
@@ -0,0 +1,21 @@
...
+    type: string
+    title: Gene ID
+  _to:
+    type: string
+    title: Gene or Phenotype ID
+  score:
+    title: Edge Score (Weight)
+    type: number
+  edge_type:
+    $ref: ../edge_type.yaml
```
ialarmedalien, line 21: Note relative `$ref`

relation_engine_server/test/test_json_validation.py
```diff
@@ -38,9 +41,16 @@
 }
 
 valid_json_loc = '/properties/params'
+test_data_dir = os_path.join('/app', 'relation_engine_server', 'test', 'data', 'json_validation')
 
+test_schema_list = [
+    ['schema', test_schema],
+    ['schema_file', os_path.join(test_data_dir, 'test_schema.json')],
+    ['schema_file', os_path.join(test_data_dir, 'test_schema.yaml')],
```
ialarmedalien, line 49: Same schema as python, JSON, and YAML.

relation_engine_server/test/test_json_validation.py
```diff
@@ -85,44 +95,243 @@ def test_non_validation_validator_errors(self):
...
+        for test_schema in test_schema_list:
+            schema_file_arg = schema_arg = test_schema[1]
+
+            if test_schema[0] == 'schema':
+                schema_file_arg = None
+            else:
+                schema_arg = None
+
+            for test in test_list:
+                test(schema_arg, schema_file_arg)
```
ialarmedalien, line 116: Run all the tests in `test_list` using a python data structure for the schema or by retrieving the schema from a JSON or YAML file.

relation_engine_server/utils/json_validation.py
```diff
@@ -114,3 +135,32 @@ def _load_json_schema(file):
             return json.load(fd)
 
         raise TypeError('Unknown file type encountered: ' + file)
+
+
+class ExtendedRefResolver(RefResolver):
+
+    def resolve_remote(self, uri):
```
ialarmedalien, line 142: This is almost entirely lifted directly from `jsonschema`

relation_engine_server/utils/json_validation.py
```diff
@@ -114,3 +135,32 @@ def _load_json_schema(file):
...
+
+        scheme = urlsplit(uri).scheme
+        # if there's no scheme, it's a local file, so prefix it with "file://"
+        if scheme == '':
+            uri = 'file://' + uri
```
ialarmedalien, line 147: Insert - assume `$ref`s in the form `/path/to/reference.json` are local files, to be accessed using the `file://` protocol

relation_engine_server/utils/json_validation.py
```diff
@@ -114,3 +135,32 @@ def _load_json_schema(file):
...
+        elif scheme in [u"http", u"https"]:
+            # Requests has support for detecting the correct encoding of
+            # json over http
+            result = requests.get(uri).json()
+        else:
+            # Otherwise, pass off to urllib and assume utf-8
+            with urlopen(uri) as url:
+                content = url.read().decode("utf-8")
+                if uri.endswith('.yaml') or uri.endswith('.yml'):
+                    result = yaml.safe_load(content)
```
ialarmedalien, line 160: Insert - read yaml files.

relation_engine_server/test/data/json_validation/invalid_date.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "invalid_date",
+  "distance": 1,
+  "date": "20200606"
+}
```
jayrbolton, line 5: This is nitpicky, don't need to change this here, but some of these files have newlines at the end and some dont. It's best to have newlines at the end of all text files so that it works well with shell tools (eg https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline)

jayrbolton: (usually the editor should insert a newline at the end of the file, without any blank line visible to the user)

ialarmedalien: Done!

relation_engine_server/utils/json_validation.py
jayrbolton: Cool! 
Detecting & loading json or yaml is definitely nice

relation_engine_server/utils/json_validation.py
jayrbolton: So with a relative reference like `my/file.json` it will make the `$ref` into `file://my/file.json`? For those to work, you would have to get the current working directory and insert that into the path. `file://` is always absolute.

If we serve these schemas over http at some point, and they are served with relative paths in the `$ref` without any `$id` being set, then we will have to set some base URIs from the client as the server URL

relation_engine_server/utils/json_validation.py
```diff
@@ -10,10 +10,17 @@
     Validator(schema).validate(obj)
```
jayrbolton, line 10: I have a feeling that down the line we will want to make this into a pip package so we can use it in other repos

spec/collections/deltaloader/delta_load_registry.yaml
```diff
@@ -5,7 +5,7 @@ schema:
   "$schema": http://json-schema.org/draft-07/schema#
   title: delta_load_registry
   type: object
-  description: Don't touch this. It's for the exlusive use of delta loaders.
+  description: Don't touch this. It's for the exclusive use of delta loaders.
```
jayrbolton, line 8: I swear this was corrected in a previous PR.....

ialarmedalien: it was in a PR on `relation_engine_spec` which I'm migrating over to this repo

relation_engine_server/utils/json_validation.py
ialarmedalien: At the moment, these refs are only followable if you give the RefResolver a file path as well as a schema -- if you just use a schema, there's no way for the ref resolver to work out where the file is.

I have written a hack that injects an `$id` field into the schemas to solve the serving over http problem.

relation_engine_server/utils/json_validation.py
```diff
@@ -60,10 +67,19 @@ def get_schema_validator(schema=None, schema_file=None, validate_at=""):
     # get the appropriate location in the schema
     validation_schema = resolve_pointer(schema, validate_at)
 
-    return Validator(validation_schema, format_checker=FormatChecker())
+    if schema_file:
+        resolver = ExtendedRefResolver(schema_file, schema)
```
jayrbolton, line 71: If this is the call that sets the base URI for the schemas, then wouldn't it make more sense to prepend the `file://` protocol to `schema_file` here?

ialarmedalien: that won't work for all ref resolution as `resolve_remote` (the overridden method from the RefResolver) works on `$ref` values, which is where the `file://` protocol needs to be added. Before calling `resolve_remote`, relative references like `../file.json` are made absolute by comparing them with the base URI, but the scheme (`file://`) doesn't get transferred over from the base URI. Thus there's still the problem of telling the RefResolver how to deal with URIs that don't have a scheme.

e.g. in the schema:

```
  $ref: ../file.yaml
```

`resolve_remote` will get `../file.yaml` as an argument

jayrbolton: Reread your comment again.. my understanding now is:
* You want to overwrite resolve_remote to handle both json & yaml
* Resolve remote does not get the base uri protocol properly passed through

ialarmedalien: Yes, that is correct. Relative paths are resolved against the base uri, but the protocol is not passed through.

jayrbolton: Okay, I updated my example to extend RefResolver to handle yaml. It still works fine with the given `file://` base URI. So I'm confused

jayrbolton: In my example, `file://` is passed through to the `resolve_remote` method. Here is what gets printed for me:

`resolve_remote uri is file:///home/j/tmp/jsonschema-resolver/resolver-test/ref1.json`

`resolve_remote uri is file:///home/j/tmp/jsonschema-resolver/resolver-test/ref2.yaml`

ialarmedalien: It may be that your example is too simple. Checkout this branch and make the following edits:

c. line 70
```
    if schema_file:
        # ADD LINE
        schema_file = 'file://' + schema_file
        resolver = ExtendedRefResolver(schema_file, schema)
    else:
        resolver = ExtendedRefResolver.from_schema(schema)
```

and further down in the file:

c. line 143

```
    def resolve_remote(self, uri):

        print({'base_uri': self.base_uri, 'current uri': uri})

        scheme = urlsplit(uri).scheme
        # if there's no scheme, it's a local file, so prefix it with "file://"
#         if scheme == '':
#             uri = 'file://' + uri
```

then run the tests

ialarmedalien: relevant test output:
```======================================================================
ERROR: test_schema_references (test_json_validation.TestJsonValidation)
Ensure referenced schemas, including those written in yaml, can be accessed.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/jsonschema/validators.py", line 774, in resolve_from_url
    document = self.store[url]
  File "/usr/local/lib/python3.7/site-packages/jsonschema/_utils.py", line 22, in __getitem__
    return self.store[self.normalize(uri)]
KeyError: '/app/relation_engine_server/test/data/schema_refs/level_1/edge_type.json'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/jsonschema/validators.py", line 777, in resolve_from_url
    document = self.resolve_remote(url)
  File "/app/relation_engine_server/utils/json_validation.py", line 160, in resolve_remote
    with urlopen(uri) as url:
  File "/usr/local/lib/python3.7/urllib/request.py", line 222, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/local/lib/python3.7/urllib/request.py", line 510, in open
    req = Request(fullurl, data)
  File "/usr/local/lib/python3.7/urllib/request.py", line 328, in __init__
    self.full_url = url
  File "/usr/local/lib/python3.7/urllib/request.py", line 354, in full_url
    self._parse()
  File "/usr/local/lib/python3.7/urllib/request.py", line 383, in _parse
    raise ValueError("unknown url type: %r" % self.full_url)
ValueError: unknown url type: '/app/relation_engine_server/test/data/schema_refs/level_1/edge_type.json'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/app/relation_engine_server/test/test_json_validation.py", line 363, in test_schema_references
    data=valid_edge_data,
  File "/app/relation_engine_server/utils/json_validation.py", line 116, in run_validator
    if validator.is_valid(data):
  File "/usr/local/lib/python3.7/site-packages/jsonschema/validators.py", line 362, in is_valid
    error = next(self.iter_errors(instance, _schema), None)
  File "/usr/local/lib/python3.7/site-packages/jsonschema/validators.py", line 328, in iter_errors
    for error in errors:
  File "/app/relation_engine_server/utils/json_validation.py", line 33, in set_defaults
    for error in validate_properties(validator, properties, instance, schema):
  File "/usr/local/lib/python3.7/site-packages/jsonschema/_validators.py", line 286, in properties
    schema_path=property,
  File "/usr/local/lib/python3.7/site-packages/jsonschema/validators.py", line 344, in descend
    for error in self.iter_errors(instance, schema):
  File "/usr/local/lib/python3.7/site-packages/jsonschema/validators.py", line 328, in iter_errors
    for error in errors:
  File "/usr/local/lib/python3.7/site-packages/jsonschema/_validators.py", line 259, in ref
    scope, resolved = validator.resolver.resolve(ref)
  File "/usr/local/lib/python3.7/site-packages/jsonschema/validators.py", line 766, in resolve
    return url, self._remote_cache(url)
  File "/usr/local/lib/python3.7/site-packages/jsonschema/validators.py", line 779, in resolve_from_url
    raise exceptions.RefResolutionError(exc)
jsonschema.exceptions.RefResolutionError: unknown url type: '/app/relation_engine_server/test/data/schema_refs/level_1/edge_type.json'

----------------------------------------------------------------------
```
i.e. you still get the paths without the protocol

jayrbolton: Okay, I will merge to not hold you up but want to look at this a little bit more

PR #21, opened 2020-08-25
Add docker build

### LINKED ISSUE #3: [API] Github action to deploy to dockerhub
Rather than using hooks/build, we can use a github action to auto-deploy.

### LINKED ISSUE #5: Docker build version
Move from using a hard-coded build version in `scripts/local-build.sh` to using a `VERSION` env var.

ialarmedalien:
Github action to build and deploy on Dockerhub.

Tested by using my own docker hub repo to create a build: [dockerhub build](https://hub.docker.com/repository/docker/ialarmedalien/relation_engine_api), [github action results](https://github.com/kbase/relation_engine/actions/runs/223994272)

Will require the appropriate docker repo credentials to perform the upload.

Closes #3 
Closes #5


.github/workflows/run_tests.yaml
```diff
@@ -16,3 +16,34 @@ jobs:
         docker-compose build
         docker-compose run re_api sh scripts/run_tests.sh
         docker-compose down --remove-orphans
+
+  docker_build_and_push:
+    runs-on: ubuntu-latest
+    needs: run_tests
+    if: (github.ref == 'refs/heads/develop' || github.ref == 'refs/heads/master') && github.event_name == 'push' && !contains(github.event.head_commit.message, 'skip_docker_build')
```
ialarmedalien, line 23: runs on pushes (i.e. merges) to `develop` or `master`
to skip the upload, include `skip_docker_build` in the commit message

.github/workflows/run_tests.yaml
```diff
@@ -16,3 +16,34 @@ jobs:
...
+      run: |
+        mkdir -p .target
+        cp VERSION .target/TAG_NAME
+
+    - name: set env vars
+      shell: bash
+      run: |
+        echo ::set-env name=DATE::$(date -u +"%Y-%m-%dT%H:%M:%SZ")
+        echo ::set-env name=BRANCH::$(git symbolic-ref --short HEAD)
+        echo ::set-env name=COMMIT::$(git rev-parse --short HEAD)
```
ialarmedalien, line 39: set env vars

jayrbolton: Can you not just call scripts/local-build here? That way we don't have redundant docker deploy code

.github/workflows/run_tests.yaml
jayrbolton: nice

.github/workflows/run_tests.yaml
jayrbolton: Clarification: I do still want a local script that can be run for quickly deploying to CI for debugging, without having to push to develop or master. I do this a lot. It seems like we could reuse that script in the github action.

jayrbolton: Also, we can move the code from `hooks/build` directly into `scripts/local-build.sh` (maybe we can call it `scripts/docker_deploy` or `scripts/docker_release`) instead. We can remove the "hooks" directory as that is specific to dockerhub.

ialarmedalien: Consolidated all the local build stuff into `scripts/local-build.sh` and removed the hooks folder. The GitHub action has to be set up as it is because all the communication with dockerhub, including passing over the docker username and password, happens as part of the action. Just using the local-build script would mean introducing potential security problems as that info would need to be exposed.

PR #23, opened 2020-08-28
Add SPEC_TEST_READY env var

ialarmedalien:
Add an environment variable that indicates when specs from the `/app/spec` directory (i.e. whatever is in `spec` when the docker image is created) have been loaded into ArangoDB. This ensures that all the tests in `spec/`, `relation_engine_server/`, and `importers/` can be run without needing to be in any special order. The tests in `client_src` are (still) a special case as they need a certain `PYTHONPATH` env var to be set.

- [x] I updated the README.md docs to reflect this change. - N/A

### DIFF
--- a/importers/test/test_djornl_parser_integration.py
+++ b/importers/test/test_djornl_parser_integration.py
@@ -1,0 +1,27 @@
+"""
+Tests for the DJORNL Parser
+
+At the present time, this just ensures that the files are parsed correctly;
+it does not check data loading into the db.
+"""
+import unittest
+import os
+
+from importers.djornl.parser import DJORNL_Parser
+from spec.test.helpers import modified_environ, check_spec_test_env
+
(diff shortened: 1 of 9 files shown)


importers/test/test_djornl_parser_integration.py
```diff
@@ -0,0 +1,27 @@
...
+    @classmethod
+    def setUpClass(cls):
+        check_spec_test_env()
+
+    def test_the_full_shebang(self):
+
+        with modified_environ(RES_ROOT_DATA_PATH=os.path.join(_TEST_DIR, 'djornl', 'test_data')):
+            parser = DJORNL_Parser()
+            parser.load_data()
+            self.assertEqual(True, parser.load_data())
```
jayrbolton, line 27: This can be `assertTrue(parser.load_data())`

Issue #11, opened 2020-08-04
Allow use of `additionalProperties` in stored query schemas


ialarmedalien:
This will require cleaning up the injection of `ws_ids` in `run_query` of `api_v1.py` -- see https://github.com/kbase/relation_engine/blob/develop/relation_engine_server/api_versions/api_v1.py#L53

jayrbolton:
I mentioned in slack that I think this can be achieved most simply by removing this line: https://github.com/kbase/relation_engine/blob/develop/relation_engine_server/api_versions/api_v1.py#L63

That line is unnecessary in the logic of what's going on there, because line 76 will always overwrite a "ws_ids" param before sending the query to arango, and after validating with the json schema

PR #26, opened 2020-09-01
Ensure that DJORNL queries do not have additional properties

ialarmedalien:
Disallow any additional parameters in the DJORNL queries.

- Remove injected `ws_ids` param unless it is required by the query
- Add "additionalProperties" to DJORNL stored queries and add tests
- Add generic `distance.yaml` file for queries with a distance param

Closes https://github.com/kbase/relation_engine/issues/11

- [x] I updated the README.md docs to reflect this change. -- N/A

### DIFF
--- a/relation_engine_server/api_versions/api_v1.py
+++ b/relation_engine_server/api_versions/api_v1.py
@@ -71,10 +71,8 @@ def run_query():
      - public stored queries (these have access controls within them based on params)
     """
     json_body = parse_json.get_json_body() or {}
-    # Don't allow the user to set the special 'ws_ids' field
-    json_body['ws_ids'] = []
-    auth_token = auth.get_auth_header()
     # Fetch any authorized workspace IDs using a KBase auth token, if present
+    auth_token = auth.get_auth_header()
     ws_ids = auth.get_workspace_ids(auth_token)
     # fetch number of documents to return
     batch_size = int(flask.request.args.get('batch_size', 10000))
@@ -99,21 +97,13 @@ def run_query():
         query_name = flask.request.args.get('stored_query') or flask.request.args.get('view')
         stored_query = spec_loader.get_stored_query(query_name)
 
-        has_ws_ids = False
-        if 'ws_ids' in stored_query['query']:
-            json_body['ws_ids'] = ws_ids
-            has_ws_ids = True
-        else:
-            del json_body['ws_ids']
-
-        stored_query_source = _preprocess_stored_query(
-            stored_query['query'], stored_query, has_ws_ids
-        )
-
         if 'params' in stored_query:
             # Validate the user params for the query
             stored_query_path = spec_loader.get_stored_query(query_name, path_only=True)
             run_validator(schema_file=stored_query_path, data=json_body, validate_at='/params')
+
+        stored_query_source = _preprocess_stored_query(stored_query['query'], stored_query)
+        json_body['ws_ids'] = ws_ids
 
         resp_body = arango_client.run_query(query_text=stored_query_source,
                                             bind_vars=json_body,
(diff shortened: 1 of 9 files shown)


relation_engine_server/api_versions/api_v1.py
```diff
@@ -90,22 +91,36 @@ def run_query():
...
         query_name = flask.request.args.get('stored_query') or flask.request.args.get('view')
         stored_query = spec_loader.get_stored_query(query_name)
-        stored_query_source = _preprocess_stored_query(stored_query['query'], stored_query)
+
+        has_ws_ids = False
+        if 'ws_ids' in stored_query['query']:
+            json_body['ws_ids'] = ws_ids
+            has_ws_ids = True
+        else:
+            del json_body['ws_ids']
```
ialarmedalien, line 107: If the stored query doesn't have `ws_ids` in it, the additional `ws_ids` param will not be needed.

relation_engine_server/api_versions/api_v1.py
```diff
@@ -90,22 +91,36 @@ def run_query():
...
+        stored_query_source = _preprocess_stored_query(
+            stored_query['query'], stored_query, has_ws_ids
+        )
+
         if 'params' in stored_query:
             # Validate the user params for the query
-            run_validator(schema=stored_query['params'], data=json_body)
-        json_body['ws_ids'] = ws_ids
+            stored_query_path = spec_loader.get_stored_query(query_name, path_only=True)
+            run_validator(schema_file=stored_query_path, data=json_body, validate_at='/params')
```
ialarmedalien, line 116: Use the path, so that relative references in the files work.

relation_engine_server/api_versions/api_v1.py
jayrbolton: As I mentioned in [this issue](https://github.com/kbase/relation_engine/issues/11), I think a simpler way to achieve the same thing with less code is to just remove line 75 here: https://github.com/kbase/relation_engine/blob/develop/relation_engine_server/api_versions/api_v1.py#L75

jayrbolton:
Done reviewing, one comment above

relation_engine_server/api_versions/api_v1.py
ialarmedalien: If the query doesn't use `ws_ids`, it is just potentially slowing things down by retrieving the list of IDs and adding them to the query. The cleanest solution would be to check whether the query/stored query mentions `ws_ids`, and only if it does, fetch the IDs from the workspace to add to the request object and alter the AQL to include the ` LET ws_ids = @ws_ids ` line.

I've reverted things to how they were, with the `ws_ids` injection, but moved the setting of `ws_ids` to after JSON validation.

jayrbolton: Avoiding an extra workspace request when the query does not use ws_ids is a good point, I agree that would be a good additional benefit

relation_engine_server/api_versions/api_v1.py
```diff
@@ -71,14 +71,13 @@ def run_query():
...
-    # Don't allow the user to set the special 'ws_ids' field
-    json_body['ws_ids'] = []
-    auth_token = auth.get_auth_header()
     # Fetch any authorized workspace IDs using a KBase auth token, if present
+    auth_token = auth.get_auth_header()
     ws_ids = auth.get_workspace_ids(auth_token)
     # fetch number of documents to return
     batch_size = int(flask.request.args.get('batch_size', 10000))
     full_count = flask.request.args.get('full_count', False)
+
```
jayrbolton, line 80: You don't have to revert, but i prefer not having all these additional newlines in the function body

relation_engine_server/api_versions/api_v1.py
jayrbolton: I'm seeing the current version of the code doesn't try to avoid the workspace request, and also doesn't remove line 75. Maybe you haven't pushed yet?

ialarmedalien: OK, I have edited it to only add `ws_ids` and pull the list from the workspace when `ws_ids` is mentioned in the query.

relation_engine_server/api_versions/api_v1.py
ialarmedalien: I find adding newlines aids readability -- otherwise it's just a bunch of lines with the same indentation and there's added cognitive load in understanding the code before you can edit it.

PR #30, opened 2020-09-03
merge silva spec

n1mus:


spec/collections/silva/silva_taxon.yaml
```diff
@@ -0,0 +1,53 @@
...
+    release:
+      type: number
+      description: SILVA release number, primarily for taxon nodes
+      examples: [138.1, 138, 132, 128, 123.1, 123, 119.1, 119]
+    sequence:
+      type: string
+      description: rRNA sequence for sequence nodes
+    dataset:
+      type: array
+      items: string
```
ialarmedalien, line 49: should be
```
items:
  type: string
```

spec/collections/silva/silva_taxon.yaml
```diff
@@ -0,0 +1,53 @@
...
+      description: The datasets that a sequence node is from. Composed of 'parc', 'ref', and 
+        'nr99', corresponding to the Parc, Ref and Ref NR99 datasets, respectively. 
+        Parc > Ref > NR99, with > denoting superset.
+      examples: [['parc'], ['parc', 'ref'], ['parc', 'ref', 'nr99']]
```
ialarmedalien, line 53: If there are only a set number of options, you can use an `enum` field to specify those options, or use the `anyOf` keyword together with a list of potential values -- see https://json-schema.org/understanding-json-schema/reference/combining.html.

spec/collections/silva/silva_taxon.yaml
```diff
@@ -0,0 +1,53 @@
...
+      examples: ['Ewamiania TS0513', 'Methyloligellaceae', 'BCP clade', 'uncultured', 
+        'Bordetella pertussis']
+    rank:
+      type: string
+      description: SILVA's taxonomic rank, with addition of `root_rank` and `sequence` for
+        root and sequence nodes, respectively
+      examples: ['superfamily', 'subphylum', 'subfamily', 'phylum', 'order', 'major_clade', 
+        'infraclass', 'suborder', 'family', 'superkingdom', 'domain', 'superphylum', 'superorder', 
+        'superclass', 'infraphylum', 'subclass', 'genus', 'class', 'kingdom', 'subkingdom', 
+        'root_rank', 'sequence']
```
ialarmedalien, line 39: As with the `dataset` option, you can use an `enum` field here if you only have a certain range of valid values.

ialarmedalien (dismissed):
A couple of suggestions for improving the schema and an error to correct.

.gitignore
```diff
@@ -12,3 +12,4 @@
 coverage_report/
 .coverage
 *.egg-info/
+.*.sw[a-z]
```
jayrbolton, line 15: If this is for vim swap files, let's remove it from here and you can put it in your global gitignore. Info about this here: https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

n1mus: oh ... thanks. i will include that next push

spec/test/collections/test_silva.py
```diff
@@ -0,0 +1,266 @@
+from yaml import safe_load
+from jsonschema import validate
```
ialarmedalien, line 2: We have a custom validator that we use in the RE repo; it will take read the spec files from the supplied path, so you just need to change your tests to use that interface.
The validator is at `relation_engine_server/utils/json_validation.py`, and you can see some examples of running validation in the tests at `relation_engine_server/test/test_json_validation.py` or `spec/test/collections/test_djornl.py`.

spec/test/collections/test_silva.py
```diff
@@ -0,0 +1,266 @@
...
+                    'from': '2',
+                    'to': 0,
+                },
+                "0 is not of type 'string'",
+            )
+        ]
+
+    def _test_type(self, schema, valid, invalid_errors):
+        for inst in valid:
+            validate(inst, schema=schema)
```
ialarmedalien, line 250: with the new validator, this will become
```
  run_validator(schema_file='/path/to/schema', data=inst)
```

spec/test/collections/test_silva.py
```diff
@@ -0,0 +1,266 @@
...
+            with self.assertRaises(ValidationError) as cm:
+                validate(inst, schema=schema)
+            msg = str(cm.exception).split('\n')[0]
+
+            print(msg)
+            self.assertTrue(msg == err_expected, '`%s` vs `%s`' % (msg, err_expected))
+
+    def test(self):
+        self._test_type(self.schema_node, self.nodes_valid, self.nodes_invalid_errors)
+        self._test_type(self.schema_edge, self.edges_valid, self.edges_invalid_errors)
```
ialarmedalien, line 262: If you're running a lot of similar tests, it's worth using `unittest`'s `subTest()` method; it means that if one of the tests fails, it will continue testing the next input, rather than just dying. See the docs: https://docs.python.org/3.6/library/unittest.html#distinguishing-test-iterations-using-subtests

spec/test/collections/test_silva.py
```diff
@@ -0,0 +1,267 @@
...
+    All information is from SILVA (arb-silva.de)
+    See their documentation for more details
+    '''
+
+    @classmethod
+    def setUpClass(cls):
+        with open(node_yaml_flpth) as fh:
+            cls.schema_node = safe_load(fh)['schema']
+        with open(edge_yaml_flpth) as fh:
+            cls.schema_edge = safe_load(fh)['schema']
```
ialarmedalien, line 26: The validator will do all this for you. Specify the schema file and where in the schema you want to validate against:

````py
run_validator(
   schema_file='/path/to/schema/file.yaml',
   data={'this': 'that'},
   validate_at='/schema',
)
````

If you're going to be doing a lot of validation against the same schema, you can load it up in advance using [`get_schema_validator`](https://github.com/kbase/relation_engine/blob/develop/relation_engine_server/utils/json_validation.py#L42) to get a validator instance:
````py
validator = get_schema_validator(
   schema_file='/path/to/schema/file.yaml',
   validate_at='/schema'
)
validator.is_valid(data_structure)
# alternative way to call the function
validator.validate(data_structure_2)
````

File: spec/test/collections/test_silva.py [hunk 9]
```diff
@@ -0,0 +1,267 @@
...
+                    'from': '2',
+                    'to': 0,
+                },
+                "0 is not of type 'string'",
+            )
+        ]
+
+    def _test_type(self, schema, valid, invalid_errors):
+        for inst in valid:
+            run_validator(schema=schema, data=inst)
```
ialarmedalien, line 251: you can add a `with self.subTest()` in here too

n1mus: thanks

ialarmedalien (approved):
LGTM!

Issue #29, opened 2020-09-03
Ensure defaults are filled in for stored queries


ialarmedalien:
Stored queries do not currently fill in the default values if they are not supplied as part of the params. Make it so!

jayrbolton:
Wait really? But we that customization here:
https://github.com/kbase/relation_engine/blob/develop/relation_engine_server/utils/json_validation.py#L26
And you have tests for it here:
https://github.com/kbase/relation_engine/blob/develop/relation_engine_server/test/test_json_validation.py#L144

ialarmedalien:
I was looking to add default values for empty lists, but it doesn't look like the python jsonschema supports it.

ialarmedalien:
OK, it looks like this is an issue with the way defaults are set, as the validator just blindly checks the subschema to see if there's a default field without first checking whether the subschema is a reference or not. 😖

PR #31, opened 2020-09-08
Spec file and validation edits and additions

ialarmedalien:
Miscellaneous edits to the spec files and code related to spec files:

- Add tests to `test_json_validation.py` for array validation
- Add `spec/test/collections/test_djornl.py` to test DJORNL collection schemas
- Refactor DJORNL node schema to use references throughout
- Add `uniqueItems` check to DJORNL stored queries
- Edit `scripts/run_tests.sh` to auto-discover all `relation_engine_server/`, `spec/`, and `importers/` tests from the base directory


relation_engine_server/test/data/json_validation/invalid_array_items.yaml
```diff
@@ -0,0 +1,6 @@
+name: invalid_array
```
jayrbolton, line 1: Why test both yaml and json for everything? I could see that making sense for the date type or some other yaml quirk, but for anything else it seems overboard.

relation_engine_server/test/test_json_validation.py
```diff
@@ -315,6 +358,74 @@ def date_format_validation(self, schema_arg, schema_file_arg):
                 data_file=file_path,
                 validate_at=valid_json_loc)
 
+    def test_array_validation(self, schema_arg=None, schema_file_arg=None):
```
jayrbolton, line 361: Why are we testing this? It makes sense to me to test something like setting defaults, since that involves our own custom code. But array validation is basic functionality covered by a third party lib.

spec/datasets/djornl/csv_node.yaml
```diff
@@ -26,10 +29,11 @@ properties:
     $ref: definitions.yaml#definitions/djornl_node/tair_curator_summary
   tair_short_description:
     $ref: definitions.yaml#definitions/djornl_node/tair_short_description
+  # comma-separated array of GO terms
   go_terms:
     type: string
     format: regex
-    pattern: "^(GO:\\d{7}, ?)*(GO:\\d{7})?$"
+    pattern: ^(GO:\d{7}, ?)*(GO:\d{7})?$
```
jayrbolton, line 36: If you want to still quote things (eg due to colons), you can use single quotes which dont require escaping backslashes. Here's a reference: https://docs.octoprint.org/en/master/configuration/yaml.html#scalars

relation_engine_server/test/test_json_validation.py
ialarmedalien: I had a weird problem with arrays not validating properly, and I couldn't find anything in the `python-jsonschema` docs or the generic `jsonschema` docs to explain what was going on. I wrote some tests to work out why things weren't working, and then edited the test input until the tests passed. Between `python-jsonschema` and `pyyaml`, I don't have much trust in the third party libs not to do something surprising or to not do something expected, so I write tests with examples of stuff I want to do to ensure it's going to work.

relation_engine_server/test/data/json_validation/invalid_array_items.yaml
ialarmedalien: The test helper I'm using automatically tests the different file types, and it takes very little time to generate JSON from YAML. May as well be consistent with all the tests.

ialarmedalien:
Will add in tests for https://github.com/kbase/relation_engine/issues/29 to this branch.

relation_engine_server/test/test_json_validation.py
jayrbolton: But I would argue that all this additional code has maintenance and runtime costs, and that if we decide that we want to write test coverage for library dependencies, then we have to cover dozens of third party libraries in our dependency tree, which nobody would ever take on. I am sure that the json-schema library has coverage on array types and that we gain little to no benefit in the cost of keeping and maintaining that test code.

spec/test/djornl/results.json
```diff
@@ -230,26 +230,26 @@
         }
       },
       {
-        "params": { "keys": ["Mary Poppins"], "distance": 0, "edge_types": [] },
```
ialarmedalien, line 233: defaults problem fixed / worked around

relation_engine_server/test/test_json_validation.py
ialarmedalien: I don't suggest that we write tests for every single third-party library that we use, but I do think it is useful to have some jsonschema tests that provide specific use case examples where the existing JSONSchema documentation is lacking or where there's some unexpected behaviour to be elucidated. I frequently look through tests to find usage examples if the docs aren't up to snuff. I've added a number of JSON validation tests specifically because the existing documentation of some feature or other was poor, I couldn't find any examples, and the best way to investigate what was going on was to create test cases and check that they worked.

In the specific case here, I've been trying to find out why array defaults weren't being filled in, whether it was a bug in `python-jsonschema`, some weird `pyyaml` thing, a problem with the schema itself, etc. It turns out that it's a shortcoming of the way that defaults are implemented -- not directly a jsonschema problem, but one due to our copying the simple default setting method in the jsonschema docs. I've written a test that directly illustrates the problem, as well as similar cases that do work.

spec/stored_queries/djornl/djornl_fetch_all.yaml
```diff
@@ -5,7 +5,17 @@ params:
...
+      description: Edge types to filter on
+      type: array
+      items:
+        $ref: ../../datasets/djornl/edge_type.yaml
+      default: []
+      uniqueItems: true
+      examples:
+        - ['AraNetv2-HT_high-throughput-ppi', 'AraNetv2-LC_lit-curated-ppi']
+        - ['AraGWAS-Phenotype_Associations']
+        - []
```
ialarmedalien, line 18: workaround for defaults problem

relation_engine_server/test/test_json_validation.py
```diff
@@ -315,6 +361,76 @@ def date_format_validation(self, schema_arg, schema_file_arg):
...
+                        'distance': 1,
+                        'fruits': []
+                    }
+                }
+            },
+            {
+                # N.b. the default does not get populated in this case!
+                # This is a change from the expected functionality
+                'fruits': fruits_array_ref,
+                'name': 'using fruits_array.yaml -- the array is a ref',
```
ialarmedalien, line 401: see https://github.com/kbase/relation_engine/issues/29

Issue #22, opened 2020-08-28
Patch potential security hole in spec_loader.py


ialarmedalien:
`relation_engine_server/utils/spec_loader.py` takes user input and uses it to fetch files from the `spec` directory. Relative paths (using `../../`) are allowed by the spec loader, with no protection against paths that aren't in the `spec` directory. Ensure that the spec loader does not permit files outside the `spec` directory (or whatever is configured as the place where spec files are stored) to be fetched.

PR #32, opened 2020-09-08
Patch security hole in spec_loader.py

ialarmedalien:
- patch security hole in `spec_loader.py` and add tests (closes https://github.com/kbase/relation_engine/issues/22)
- add "get_view" method to `spec_loader.py` and add a view to the `sample_spec_release` test set
- add `sample_spec_release` dir (expanded version of the test spec archive, `spec.tar.gz`)
- add instructions for updating the test spec archive


relation_engine_server/test/test_spec_loader.py
```diff
@@ -7,20 +7,37 @@
...
+
+        config = get_config()
+        cls.repo_path = config['spec_paths']['repo']
+        for key in config['spec_paths'].keys():
+            if cls.repo_path in config['spec_paths'][key]:
+                config['spec_paths'][key] = config['spec_paths'][key].replace(
+                    cls.repo_path,
+                    cls.test_spec_dir
+                )
+        cls.config = config
```
ialarmedalien, line 27: Edit `config` settings to point to the `sample_spec_release` dir for the duration of this test.

relation_engine_server/test/test_spec_loader.py
```diff
@@ -162,3 +185,22 @@ def test_get_schema_for_doc(self):
...
+        # this query is OK as the file is still in the spec repo
+        path_in_spec_repo = '../../../../../**/fetch_test_vertex'
+        result = spec_loader.get_schema('stored_queries', path_in_spec_repo, path_only=True)
+        self.assertEqual(
+            result,
+            os_path.join(self.test_spec_dir, 'stored_queries', 'test', 'fetch_test_vertex.yaml')
+        )
+
+        # this matches a file in one of the other test data dirs => should throw an error
+        path_outside_spec_repo = '../../../../data/collections/test_node'
```
ialarmedalien, line 203: Prior to the fixes, this would retrieve a file. 😱

relation_engine_server/test/test_spec_loader.py
```diff
@@ -125,6 +144,10 @@ def test_get_schemas_of_various_types(self):
                 'schema_type_names': ['stored_query', 'stored_queries'],
                 'example': 'ncbi_fetch_taxon',
             },
+            {
+                'schema_type_names': ['view', 'views'],
+                'example': 'test_vertices',
+            }
```
ialarmedalien, line 150: add in tests for retrieving views

relation_engine_server/test/spec_release/README.md
```diff
@@ -1,8 +1,22 @@
+## Test Spec Release
 
-The file in this directory, `spec.tar.gz` is a cached release of the `relation_engine_spec` repo, found here:
+`sample_spec_release`, and the corresponding archive, `spec.tar.gz`, contain a set of sample schema files suitable for use in tests.
```
jayrbolton, line 3: Seems like we should remove `spec.tar.gz`? That was only there before because the repos were separated and served as a cache of the specs that we don't have to fetch over the network, which seems moot now

relation_engine_server/utils/spec_loader.py
```diff
@@ -74,11 +74,18 @@ def get_schema(schema_type, name, path_only=False):
...
 
-    try:
-        path = _find_paths(_CONF['spec_paths'][schema_search_type], name + '.yaml')[0]
-    except IndexError:
+    yaml_paths = _find_paths(_CONF['spec_paths'][schema_search_type], f'{name}.yaml')
+    json_paths = _find_paths(_CONF['spec_paths'][schema_search_type], f'{name}.json')
+    # ensure we're using the canonical path and that all paths are unique
+    # we are only interested in paths that are in the designated spec repo
+    repo_path = os.path.abspath(_CONF['spec_paths']['repo'])
+    all_paths = [p for p in set(os.path.abspath(path) for path in yaml_paths + json_paths) if repo_path in p]
```
jayrbolton, line 82: Using `p.startswith(repo_path)` I think would be more accurate

I would avoid a nested list comprehension like this to help with readability

relation_engine_server/test/spec_release/README.md
ialarmedalien: I'd keep it, as it is exercising the `pull_spec.py` code, which would not be run otherwise; ideally it could be used to create some tests for `pull_spec.py`, too, since we don't have anything that checks the archive expansion/loading explicitly.

relation_engine_server/utils/spec_loader.py
```diff
@@ -112,6 +119,11 @@ def get_stored_query_names():
     return get_names('stored_queries')
 
 
+def get_view_names():
+    """Return an array of all stored queries base names."""
```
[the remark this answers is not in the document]
ialarmedalien: Yeah, copy-pasta error. Thanks!

relation_engine_server/utils/spec_loader.py
```diff
@@ -133,6 +145,11 @@ def get_stored_query(name, path_only=False):
     return get_schema('stored_query', name, path_only)
 
 
+def get_view(name, path_only=False):
+    """Get AQL content or file path for a specific stored query. Throws an error if nonexistent."""
```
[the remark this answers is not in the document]
ialarmedalien: copy-pasta confusion

relation_engine_server/utils/spec_loader.py
ialarmedalien: OK, have used `p.startswith` and split out the list comprehension. 👍

relation_engine_server/test/spec_release/README.md
jayrbolton: Ok, since it is just used to test the pull_spec module, then it might be a good update to automate its creation in those tests. Python has a builtin `tarfile` lib that supports gzipping, so it should only be a few lines in the test. We can do it in a future update.

PR #33, opened 2020-09-08
Improve output of spec/validate.py to give more detail on errors

ialarmedalien:
Change the output of validate.py to list validation errors at the end of output if there are issues (avoiding having to scroll back up through a long list of OK files).

- [x] I updated the README.md docs to reflect this change. - N/A


jayrbolton (approved):
:+1:

PR #38, opened 2020-09-18
Fix/clean docker build script

jayrbolton:
* Added a missing backslash character in the docker build script
* Renamed the script so it's more consistent with other core services (and made it executable)
* Image tags use the semantic version of the app, not the branch name


ialarmedalien (approved):
LGTM

PR #39, opened 2020-09-18
Fix wait for arango

jayrbolton:
On CI, we use an auth account that is not able to access /cluster/health. Switching it back for now to request the arango root endpoint without auth

Branched from #38


jayrbolton:
Looking at tests

jayrbolton:
This got messed up, closing

PR #41, opened 2020-09-18
Fix spec paths

jayrbolton:
Update things to work with the new repo directory structure. 

Removed extra code for simplification's sake

Auto-generate the test spec tarball from the run_test.sh script.


relation_engine_server/test/test_spec_loader.py
```diff
@@ -16,10 +16,10 @@ class TestSpecLoader(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         cls.test_dir = os_path.join('/app', 'relation_engine_server', 'test')
-        cls.test_spec_dir = os_path.join(cls.test_dir, 'spec_release', 'sample_spec_release')
+        cls.test_spec_dir = os_path.join(cls.test_dir, 'spec_release', 'sample_spec_release', 'spec')
 
         cls.config = get_config()
-        cls.repo_path = cls.config['spec_paths']['repo']
+        cls.repo_path = cls.config['spec_paths']['root']
```
ialarmedalien, line 22: 👍

spec/test/collections/test_djornl.py
```diff
@@ -58,7 +58,7 @@ def test_node(self, query_name=None, test_data=None):
             {
                 'data': {'_key': 'ABCDE', 'clusters': ['GO:0003700', 'GO:0003700']},
                 'valid': False,
-                'error': "\['GO:0003700', 'GO:0003700'\] has non-unique elements"
+                'error': "\\['GO:0003700', 'GO:0003700'\\] has non-unique elements"
```
ialarmedalien, line 61: did this just start throwing errors? I wonder why it didn't before.

ialarmedalien (approved):
👍

spec/test/collections/test_djornl.py
jayrbolton: I don't think there's an actual problem in runtime, but my editor's linter had a warning or error. The warning does make sense, because within double quotes you have escape characters, and `\[` is not a valid escape character. Python probably handles this by just ignoring the error and making it a literal.

PR #42, opened 2020-09-21
Remove travis

jayrbolton:
Since we have a working github workflow, I don't think we need Travis anymore, but maybe @ialarmedalien kept it in for a reason.


ialarmedalien:
I have no attachment to travis - just kept it in in case you wanted it! :)

PR #51, opened 2020-10-06
Update edge types for networks from Jacobson Lab

izaakm:


jayrbolton (approved):
Looks good and non-breaking. I don't have comments on the content.

spec/datasets/djornl/edge_type.yaml
```diff
@@ -23,3 +23,32 @@ oneOf:
...
+  - const: AtPIN_PPI
+    title: AtPIN PPI
+    description: The interactions database includes all interactions present at the Arabidopsis thaliana Protein Interactome Database, the Predicted Interactome for Arabidopsis, Arabidopsis protein-protein interaction data curated from the literature by TAIR curators, BIOGRID and IntAct. https://atpin.bioinfoguy.net/cgi-bin/atpin.pl
+      
+  - const: mentha-AT-3702-040319_PPI
+    title: Mentha AT 3702 040319 PPI
+    description: Mentha archives evidence collected from different sources and presents these data in a complete and comprehensive way. Its data comes from manually curated protein-protein interaction databases that have adhered to the IMEx consortium and assigns to each interaction a reliability score that takes into account all the supporting evidence.  https://mentha.uniroma2.it/about.php
+
+  - const: aranetv2-subnet-AT-LC_PPI
+    title: AraNetv2 subnet AT-LC PPI
```
ialarmedalien, line 40: Is this different from the existing network `AraNetv2-LC_lit-curated-ppi`?

spec/datasets/djornl/edge_type.yaml
```diff
@@ -23,3 +23,32 @@ oneOf:
...
+    description: Literature curated PPI of A. thaliana. https://www.inetbio.org/aranet/dl.php?f=AT-LC
+
+  - const: Wu2016-s015-Gene-to-Metab_GeneToPhenotype
+    title: Wu2016 s015 Gene-to-Metab GeneToPhenotype
+    description: GWAS hits ftom a Gene-to-Metaboiltes GWAS. Phenotypes (metabolites) have a unique ID from the Wu 2016 study and need to be given our own UID for future use.
+
+  - const: AraGWAS-subnet-permsig-geni_GeneToPhenotype
```
ialarmedalien, line 47: Is this different to the existing network `AraGWAS-Phenotype_Associations`? What does "permsig geni" mean?

spec/datasets/djornl/edge_type.yaml
```diff
@@ -23,3 +23,32 @@ oneOf:
...
+    title: Mentha AT 3702 040319 PPI
+    description: Mentha archives evidence collected from different sources and presents these data in a complete and comprehensive way. Its data comes from manually curated protein-protein interaction databases that have adhered to the IMEx consortium and assigns to each interaction a reliability score that takes into account all the supporting evidence.  https://mentha.uniroma2.it/about.php
+
+  - const: aranetv2-subnet-AT-LC_PPI
+    title: AraNetv2 subnet AT-LC PPI
+    description: Literature curated PPI of A. thaliana. https://www.inetbio.org/aranet/dl.php?f=AT-LC
+
+  - const: Wu2016-s015-Gene-to-Metab_GeneToPhenotype
+    title: Wu2016 s015 Gene-to-Metab GeneToPhenotype
+    description: GWAS hits ftom a Gene-to-Metaboiltes GWAS. Phenotypes (metabolites) have a unique ID from the Wu 2016 study and need to be given our own UID for future use.
```
ialarmedalien, line 45: typos: 
```
description: GWAS hits from a Gene-to-Metabolites GWAS. Phenotypes (metabolites) have a unique ID from the Wu 2016 study and need to be given our own UID for future use.
```

Have you got a PMID or DOI for the Wu 2016 study?

ialarmedalien (commented):
Looking through the new entries, I think we could rename the existing edges to put the type of connection first, followed by the protocol / data source / etc.

current edges:
- `phenotype-association_AraGWAS` (was `AraGWAS-Phenotype_Associations`): GWAS associations from Ara 1001 Genomes DB
- `pairwise-gene-coexpression_AraNet_v2` (was `AraNetv2-CX_pairwise-gene-coexpression`): pairwise co-expr from AraNet v2
- `domain-co-occurrence_AraNet_v2` (was `AraNetv2-DC_domain-co-occurrence`)
- `protein-protein-interaction_high-throughput_AraNet_v2` (was `AraNetv2-HT_high-throughput-ppi`)
- `protein-protein-interaction_literature-curated_AraNet_v2` (was `AraNetv2-LC_lit-curated-ppi`)

new edges:
- `protein-protein-interaction_biogrid_date/release?` (BIOGRID ORGANISM Arabidopsis thaliana Columbia 3.5.188 tab3 PPI) -- 
- `protein-protein-interaction_AtPIN` (AtPIN PPI)
- `protein-protein-interaction_Mentha_A_thaliana_3702_040319` (Mentha AT 3702 040319 PPI)
- `protein-protein-interaction_literature_curated_AraNet_v2_subnet` (AraNetv2 subnet AT-LC PPI -- if it is different from the other AraNetv2 entry above)
- `phenotype-association_GWAS_gene_to_metabolite???` (Wu2016 s015 Gene-to-Metab GeneToPhenotype) -- maybe add in the DOI of the Wu study?
- `phenotype-association_AraGWAS_subnet_permsig_geni` (AraGWAS subnet permsig geni GeneToPhenotype - if different from the existing AraGWAS entry)
- `transcription-factor-regulatory-interaction_literature_curated_ATRM_01082020` (ATRM TF to Target LitCurated 01082020 TranscriptionFactorToGene)

spec/datasets/djornl/edge_type.yaml
izaakm: The two networks should be the same, or at least derived from the same source. The only difference might be if different filtering had been applied to either network.

spec/datasets/djornl/edge_type.yaml
izaakm: I believe this is a sub-network of `AraGWAS-Phenotype_Associations`. The notes I have for the file say ...

```Has the best GWAS hits from AraGWAS (i.e. passed a permutation-based threshold). 17,606 edges```

whereas I think the full `AraGWAS-Phenotype_associations` network is on the order of 1,000,000 edges.

spec/datasets/djornl/edge_type.yaml
izaakm: DOI for Wu 2016: 10.1371/journal.pgen.1006363

izaakm:
I think the updated edge names seem fine. My only comment would be that the edge names could be split into separate fields, so, eg, `protein-protein-interaction` would become one field, and `AtPIN` would become a second field. This might help users to identify, eg, all of the protein-protein interaction networks, and then filter according to source. I do not know, however, what the intended use for the networks is, so this example might be irrelevant.

ialarmedalien:
I think we'll probably want to add more detail to the edge types at some point, adding fields like 'interaction_type', 'source', etc. -- so the strings that are currently in the `edge_type` will act as unique identifiers.

ialarmedalien (approved):
👍

PR #52, opened 2020-10-08
DJORNL parser: detect more header errors, and do a dry run of parsing

ialarmedalien:
The last of the parser updates:

- store nodes (`node_ix`) and edges (`edge_ix`) on the parser object
- move the `store_parsed_*_data` refs out to be top level methods
- add checks for duplicated headers and incorrect headers
- return a list of errors after parsing each type of file, rather than throwing an error, and then throw an error once all data has been parsed
- add in a "dry run" option for the parser, which just reports some stats about the files (if no errors were found) or a list of errors (if there were problems)


importers/djornl/parser.py
```diff
@@ -164,19 +174,46 @@ def check_headers(self, headers, validator=None):
...
 
         # check that each required header in the schema is present in headers
         required_props = validator.schema['required']
-        return [i for i in required_props if i not in headers]
+        missing_headers = [i for i in required_props if i not in headers]
+        if missing_headers:
+            header_errs['missing'] = missing_headers
+
+        if 'additionalProperties' in validator.schema and validator.schema['additionalProperties'] is False:
+            all_props = validator.schema['properties'].keys()
```
ialarmedalien, line 211: unless `additionalProperties` is present and set to `False`, there is no such thing as an invalid header

importers/djornl/parser.py
```diff
@@ -284,13 +331,42 @@ def process_file(self, file, remap_fn, store_fn, err_list, validator=None):
         if not n_stored:
             err_list.append(f"{file['path']}: no valid data found")
 
+    def store_parsed_edge_data(self, datum):
```
ialarmedalien, line 334: This was defined under the `load_edges` method before - moving it and the other similar methods (`store_parsed_node_data`) out to be top level methods.

importers/djornl/parser.py
```diff
@@ -563,45 +634,82 @@ def save_docs(self, coll_name, docs, on_dupe='update'):
...
+        if all_errs:
+            raise RuntimeError("\n".join(all_errs))
 
-    def check_data_delta(self):
-        edge_data = self.load_edges()
-        node_data = self.load_nodes()
-        clusters = self.load_clusters()
+        if dry_run:
+            # report stats on the data that has been gathered
+            return self.summarise_dataset()
```
ialarmedalien, line 654: this replaces the existing `check_data_delta` / `check_deltas` methods

importers/djornl/parser.py
```diff
@@ -563,45 +634,82 @@ def save_docs(self, coll_name, docs, on_dupe='update'):
...
 
-        # count all edges
-        print("Dataset contains " + str(len(edge_data['edges'])) + " edges")
-        # count all nodes
-        print("Dataset contains " + str(len(all_nodes)) + " nodes")
+        return {
+            'nodes_total': len(self.node_ix.keys()),
+            'edges_total': len(self.edge_ix.keys()),
+            'nodes_in_edge': len(nodes_in_edge_ix.keys()),
+            'node_type_count': node_type_ix,
```
ialarmedalien, line 705: distribution of nodes by `node_type`

importers/djornl/parser.py
```diff
@@ -563,45 +634,82 @@ def save_docs(self, coll_name, docs, on_dupe='update'):
...
+            'edge_type_count': edge_type_ix,
```
ialarmedalien, line 706: distribution of edges by `edge_type`

importers/djornl/parser.py
```diff
@@ -563,45 +634,82 @@ def save_docs(self, coll_name, docs, on_dupe='update'):
...
+                edge_type_ix[edge['edge_type']] = 1
 
-        # count all edges
-        print("Dataset contains " + str(len(edge_data['edges'])) + " edges")
-        # count all nodes
-        print("Dataset contains " + str(len(all_nodes)) + " nodes")
+        return {
+            'nodes_total': len(self.node_ix.keys()),
+            'edges_total': len(self.edge_ix.keys()),
+            'nodes_in_edge': len(nodes_in_edge_ix.keys()),
```
ialarmedalien, line 704: nodes in the node dataset that are in an edge

.github/pull_request_template.md
```diff
@@ -1,4 +1,8 @@
 
 - [ ] I updated the README.md docs to reflect this change.
-- [ ] This is not a breaking API change OR 
-- [ ] This is a breaking API change and I have incremented the API version.
+
+For changes to the codebase:
+
+- [ ] I have written tests to cover this change.
+- [ ] This is not a breaking API change OR
+- [ ] This is a breaking API change and I have incremented the API version and added a summary to CHANGELOG.md.
```
jayrbolton, line 8: Some thoughts related to this:

I think we could add an entry in the changelog for nearly every PR.  The version we're using is really for the API, I suppose, and it might be worth having a separate version for the API vs. the spec, as they're two different release processes.

importers/djornl/parser.py
```diff
@@ -164,19 +174,46 @@ def check_headers(self, headers, validator=None):
...
+                                        returns an empty dict
         """
 
         if validator is None:
-            return []
+            return {}
+
+        header_errs = {}
+
+        all_headers = {}
```
jayrbolton, line 192: Another way to do this would be:
```py
all_headers = defaultdict(lambda: 0)
for h in headers:
    all_headers[h] += 1
```

importers/djornl/parser.py
```diff
@@ -237,12 +275,21 @@ def process_file(self, file, remap_fn, store_fn, err_list, validator=None):
...
-                f"{file['path']}: missing required headers: " + ", ".join(sorted(missing_headers))
-            )
+        header_errors = self.check_headers(cols, validator)
+        if header_errors.keys():
+            err_str = {
+                'duplicate': 'duplicate',
+                'missing': 'missing required',
+                'invalid': 'invalid additional',
+            }
+            for err_type in ['missing', 'invalid', 'duplicate']:
```
jayrbolton, line 285: Maybe `for err_type in err_str.keys()`?

jayrbolton (approved):

PR #53, opened 2020-10-08
Add specs to release image, and misc cleanup

jayrbolton:
In the published docker image for the API, I'd like the full set of specs cached in there so we can use the image in tests elsewhere (such as in index_runner).

This setup creates a tarball at /opt/spec.tar.gz and sets that as the default spec to pull. The docker-compose.yaml overrides this path and uses `/app/relation_engine_server/test/spec_release/spec.tar.gz` for our tests within the repo.


relation_engine_server/README.md
```diff
@@ -200,7 +200,7 @@ curl {root_url}/api/v1/update_specs
 
 _Query params_
 * `init_collections` - optional - boolean - defaults to true - whether to initialize any new collections in arango (also creates indexes and views)
-* `spec_url` - optional - string - the specific url of the release to download and use (as a tarball). If left blank, then the latest release from github is used (not including any pre-releases or drafts).
```
jayrbolton, line 203: this name was incorrect

relation_engine_server/api_versions/api_v1.py
```diff
@@ -170,7 +170,9 @@ def show_config():
         'kbase_endpoint': conf['kbase_endpoint'],
         'db_url': conf['db_url'],
         'db_name': conf['db_name'],
-        'spec_url': conf['spec_url']
+        'spec_repo_url': conf['spec_repo_url'],
```
jayrbolton, line 173: This just exposes these config vars in an http endpoint for debugging purposes.

docker-compose.yaml
```diff
@@ -13,10 +13,8 @@ services:
         DEVELOPMENT: 1
     ports:
       - "127.0.0.1:5000:5000"
-#    uncomment to mount local directories
-#    volumes:
-#      - ${PWD}:/app
-#      - "./spec:/spec"
+    volumes:
```
jayrbolton, line 16: I always want this mounted.. but maybe there a reason not to have it defaulted?

relation_engine_server/utils/arango_client.py
```diff
@@ -82,14 +82,16 @@ def create_collection(name, config):
         'keyOptions': {'allowUserKeys': True},
         'name': name,
         'type': collection_type,
-        'numberOfShards': num_shards
+        'numberOfShards': num_shards,
+        'waitForSync': True,
```
jayrbolton, line 86: The waiting for arangodb functionality was not working in other repos. This was one blind attempt to fix that, but didn't fix anything. It might still be useful to leave in though :shrug:

relation_engine_server/utils/config.py
```diff
@@ -38,7 +41,7 @@ def get_config():
         'db_pass': db_pass,
         'db_readonly_user': db_readonly_user,
         'db_readonly_pass': db_readonly_pass,
-        'spec_url': spec_url,
+        'spec_repo_url': spec_repo_url,
```
jayrbolton, line 44: Just wanted to make this name more explicit

relation_engine_server/utils/wait_for.py
```diff
@@ -5,66 +5,39 @@
...
-    if service_name not in service_conf:
-        raise KeyError(f'Configuration for {service_name} not found')
-
-    return {
-        'name': service_name,
-        # auth defaults to None if there is nothing set
-        'auth': service_conf[service_name].get('auth'),
-        'url': service_conf[service_name]['url'],
-    }
+    services_pending = set(service_list)
```
jayrbolton, line 16: To get it to work more reliably waiting for arangodb, I added the 'callback' option for the services, and in the process refactored things a little bit and added the 'raise_for_status' option.

scripts/run_tests.sh
```diff
@@ -9,17 +9,13 @@ set -e
...
-rm -rf /spec
-mkdir /spec
-mkdir /spec/repo
-cp -r /app/spec/* /spec/repo/
 # start server, using the specs in /spec/repo
 sh /app/scripts/start_server.sh &
 coverage erase
 # spec validation
-python -m spec.validate &&
+python -m spec.validate
```
jayrbolton, line 16: Since we have `set -e`, we shouldn't need the `&&`

scripts/run_tests.sh
```diff
@@ -9,17 +9,13 @@ set -e
 flake8 --max-complexity 15 /app
 mypy --ignore-missing-imports /app
 bandit -r /app
-rm -rf /spec
```
jayrbolton, line 12: I wasn't able to find the purpose of these lines if the tests use `sample_spec_release`

Small thing, but we can do `mkdir -p /spec/repo`

ialarmedalien (changes requested):

relation_engine_server/utils/wait_for.py
```diff
@@ -81,6 +54,32 @@ def wait_for_api():
...
+        'url': _CONF['auth_url'],
+    },
+    'workspace': {
+        'url': _CONF['workspace_url'],
+    },
+    'localhost': {
+        'url': 'http://127.0.0.1:5000',
+        'raise_for_status': True,
+    }
+}
```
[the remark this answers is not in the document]
jayrbolton: What makes the variable "magic"? I was thinking of this as const configuration data, but unfortunately python doesn't have actual constants (I would clone it to modify it in tests)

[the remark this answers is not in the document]
ialarmedalien: Config data is one of the things you're most likely to change, though, especially when writing tests. It's much easier to deal with if you encapsulate it in a method.

ialarmedalien (approved):

Issue #57, opened 2020-10-22
DJORNL parser: `force` option


ialarmedalien:
Add a `force` option that can be called from the command line or as a parameter to `load_data`. This option will load all the valid data that the parser has encountered into the db and report on erroneous data.

PR #58, opened 2020-10-27
Fixed #56: add --dry-run command line parameter.

### LINKED ISSUE #56: DJORNL parser: more `dry_run` options
- extend the `dry_run` option so that it can be called from the command line, e.g.

```
python -m importers.djornl.parser --dry_run
```

- alter the implementation of `dry_run` so that it prints dataset statistics and a list of errors, rather than dying with a list of errors without reporting stats.

- print out the stats in a more user-friendly manner

dakotablair:
This PR adds the `--dry-run` parameter to validate the input data and print
an output summary even in the presence of errors. To this end, the `--output`
parameter is also introduced to give an option to display the summary as JSON
or in a more user-friendly text format. The `--debug` parameter will repeat the
full list of errors when there are any at the end of the run. Errors are now
printed to stderr and may be suppressed using shell redirection.

### DIFF
--- a/importers/djornl/parser.py
+++ b/importers/djornl/parser.py
@@ -666,8 +666,8 @@ class DJORNL_Parser(object):
             if output['err_list']:
                 all_errs = all_errs + output['err_list']
 
-        # save the dataset, unless this is a dry run
-        if not dry_run:
+        # if there are no errors then save the dataset unless this is a dry run
+        if len(all_errs) == 0 and not dry_run:
             self.save_dataset()
 
         # report stats on the data that has been gathered
--- a/importers/djornl/parser.py
+++ b/importers/djornl/parser.py
@@ -15,14 +15,18 @@ from the command line:
 RES_ROOT_DATA_PATH=/path/to/data/dir python -m importers.djornl.parser
 
 """
-import json
-import requests
-import os
+import argparse
 import csv
+import json
+import logging
+import os
+import requests
 import yaml
 
 import importers.utils.config as config
 from relation_engine_server.utils.json_validation import run_validator, get_schema_validator
+
+LOGGER = logging.getLogger(__name__)
 
 
 class DJORNL_Parser(object):
@@ -48,7 +52,9 @@ class DJORNL_Parser(object):
 
     def _configure(self):
 
-        configuration = config.load_from_env(extra_required=['ROOT_DATA_PATH'])
+        configuration = config.load_from_env(
+            extra_required=['ROOT_DATA_PATH']
+        )
 
         # Collection name config
         configuration['node_name'] = 'djornl_node'
@@ -89,7 +95,9 @@ class DJORNL_Parser(object):
 
         if not hasattr(self, '_dataset_schema_dir'):
             dir_path = os.path.dirname(os.path.realpath(__file__))
-            self._dataset_schema_dir = os.path.join(dir_path, '../', '../', 'spec', 'datasets', 'djornl')
+            self._dataset_schema_dir = os.path.join(
+                dir_path, '../', '../', 'spec', 'datasets', 'djornl'
+            )
 
         return self._dataset_schema_dir
 
@@ -131,7 +139,7 @@ class DJORNL_Parser(object):
         '''Given a dict containing file information, instantiate the correct type of parser'''
 
         delimiter = '\t'
-        if 'file_format' in file and file['file_format'].lower() == 'csv' or file['path'].lower().endswith('.csv'):
+        if file.get('file_format', '').lower() == 'csv' or file['path'].lower().endswith('.csv'):
             delimiter = ','
         return csv.reader(fd, delimiter=delimiter)
 
@@ -176,8 +184,8 @@ class DJORNL_Parser(object):
 
         :return header_errs: (dict)     dict of header errors:
                                         'missing': required headers that are missing from the input
-                                        'invalid': additional headers that should not be in the input
-                                        'duplicate': duplicated headers (content would be overwritten)
+                                        'invalid': headers that should not be in the input
+                                        'duplicate': duplicated headers (data would be overwritten)
                                         If the list of headers supplied is valid--i.e. it
                                         contains all the fields marked as required in the validator
                                         schema--or no validator has been supplied, the method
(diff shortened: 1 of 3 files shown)


importers/djornl/parser.py
```diff
@@ -15,15 +15,19 @@
...
+import logging
 import os
-import csv
+import requests
 import yaml
 
 import importers.utils.config as config
 from relation_engine_server.utils.json_validation import run_validator, get_schema_validator
 
+LOGGER = logging.getLogger(__name__)
```
ialarmedalien, line 29: 👍

importers/djornl/parser.py
```diff
@@ -646,22 +666,18 @@ def load_data(self, dry_run=False):
             if output['err_list']:
                 all_errs = all_errs + output['err_list']
 
-        if all_errs:
-            raise RuntimeError("\n".join(all_errs))
+        # save the dataset, unless this is a dry run
+        if not dry_run:
+            self.save_dataset()
```
ialarmedalien, line 671: This is a change in functionality. The parser should die with a list of errors and _not_ upload files if there are any problems with the input; there's another ticket (#57) to add a `force` option to it that will load data even if there have been errors. I want to be able to use the parser to QC data, so having it die if it encounters any problems is important.

importers/djornl/parser.py
```diff
@@ -709,13 +725,94 @@ def summarise_dataset(self):
                 'cluster': len(node_data['cluster']),
                 'full': len(node_data['full'])
             },
+            'errors_total': len(errs),
```
ialarmedalien, line 728: 🎉

importers/djornl/parser.py
```diff
@@ -709,13 +725,94 @@ def summarise_dataset(self):
...
+            'errors': errs
         }
 
 
-if __name__ == '__main__':
+def format_summary(summary, output):
```
ialarmedalien, line 733: you could move this into the class so that it's a parser method

importers/djornl/parser.py
```diff
@@ -709,13 +725,94 @@ def summarise_dataset(self):
...
+---
+Edge Types
+{edge_types:{value_width}}
+---
+Node data available
+{summary['node_data_available']['key_only']:{value_width}} Key only
+{summary['node_data_available']['cluster']:{value_width}} Cluster
+{summary['node_data_available']['full']:{value_width}} Full
+---
+{summary.get('errors_total'):{value_width}} Errors
```
ialarmedalien, line 778: Nice!

importers/djornl/parser.py
```diff
@@ -709,13 +725,94 @@ def summarise_dataset(self):
...
+""".format(value_width)
+    return text_summary
+
+
+def main():
+    argparser = argparse.ArgumentParser(description='Load DJORNL data')
+    argparser.add_argument(
+        '--debug', action='store_true',
+        help='Print errors in summary, by default only their count is printed.'
```
ialarmedalien, line 787: I think it would make more sense to have this switch function in the opposite manner -- if the arg is supplied, only the error count is printed; if it isn't, the whole error list is printed. The aim is to maximise data quality and minimise errors, and having the whole list of errors printed out by default will increase the likelihood that those errors get fixed. A number is easy to ignore!

(I know that the current full data set has a ton of errors -- future iterations will, I hope, have far fewer!)

importers/djornl/parser.py
```diff
@@ -709,13 +725,94 @@ def summarise_dataset(self):
...
+    )
+    argparser.add_argument(
+        '--dry-run', dest='dry', action='store_true',
+        help='Perform all actions of the parser, except loading the data.'
+    )
+    argparser.add_argument(
+        '--output', default='text',
+        help='Specify the format of any output generated. (text or json)'
+    )
```
ialarmedalien, line 796: 👍

importers/djornl/parser.py
ialarmedalien: I'd also add the text/json switch to the `load_data` method, so that you can get the same capabilities running the parser in a script (e.g. `parser = DJORNL_Parser()` and then `parser.load_data(args)`) that you can from running it from the command line.

importers/djornl/parser.py
```diff
@@ -709,13 +725,94 @@ def summarise_dataset(self):
...
+    edge_types = "\n".join([(
+            f"{count:{value_width}} {edge_type_names.get(etype, etype)}"
+            .format(value_width)
+        )
+        for etype, count in summary['edge_type_count'].items()
+    ])
+    text_summary = f"""
+{summary['nodes_total']:{value_width}} Total nodes
+{summary['edges_total']:{value_width}} Total edges
+{summary['nodes_in_edge']:{value_width}} Nodes in edge
```
ialarmedalien, line 765: perhaps you could add a better text description to this one -- it represents nodes that are part of/one end of an edge

importers/djornl/parser.py
```diff
@@ -709,13 +725,94 @@ def summarise_dataset(self):
...
+    edge_type_names = dict()
+    edge_types = "\n".join([(
+            f"{count:{value_width}} {edge_type_names.get(etype, etype)}"
+            .format(value_width)
+        )
+        for etype, count in summary['edge_type_count'].items()
+    ])
+    text_summary = f"""
+{summary['nodes_total']:{value_width}} Total nodes
+{summary['edges_total']:{value_width}} Total edges
```
ialarmedalien, line 764: Probably easier to parse with the number second, i.e.

```
Total nodes: 4871
Total edges:    1
```

importers/djornl/parser.py
```diff
@@ -266,13 +274,18 @@ def process_file(self, file, remap_fn, store_fn, err_list, validator=None):
         :param validator: (Validator)   jsonschema validator object
 
         """
-        print("Parsing " + file['data_type'] + " file " + file['file_path'])
+        LOGGER.info("Parsing " + file['data_type'] + " file " + file['file_path'])
         file_parser = self.parser_gen(file)
+
+        def add_error(error):
+            LOGGER.error(error)
+            err_list.append(error)
```
ialarmedalien, line 282: The downside of doing it this way is that there will be a lot more output from the parser, with more repetition. If there's a whole ton of errors, you're going to have to scroll or search back in the terminal to try to find where the file summary starts. If you've done several runs of the parser and forgot to clear your terminal, this starts becoming a PITA.

ialarmedalien (changes requested):
Revert change in functionality.

PR #59, opened 2020-11-19
Add --dry-run command line parameter.

dakotablair:
This PR fixes issue #56 by adding the --dry-run parameter to validate the input data and print an output summary even in the presence of errors. To this end, the --output parameter is also introduced to give an option to display the summary as JSON or in a more user-friendly text format. Data should not be loaded if there are errors or if this command is invoked with --dry-run.

### DIFF
--- a/importers/djornl/parser.py
+++ b/importers/djornl/parser.py
@@ -15,10 +15,11 @@ from the command line:
 RES_ROOT_DATA_PATH=/path/to/data/dir python -m importers.djornl.parser
 
 """
-import json
-import requests
-import os
+import argparse
 import csv
+import json
+import os
+import requests
 import yaml
 
 import importers.utils.config as config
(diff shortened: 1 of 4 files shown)


importers/djornl/parser.py
```diff
@@ -276,11 +272,16 @@ def process_file(self, file, remap_fn, store_fn, err_list, validator=None):
         """
         print("Parsing " + file["data_type"] + " file " + file["file_path"])
         file_parser = self.parser_gen(file)
+
+        def add_error(error):
+            print(error)
+            err_list.append(error)
```
ialarmedalien, line 278: this will result in errors getting printed out twice, since they're already printed in the summary section

dakotablair: Yes, this is a vestige of the other code. I have removed the print in the commit below.

ialarmedalien (approved):
Looks great!

PR #60, opened 2020-11-19
Implement edge directionality in DJORNL graph

### LINKED ISSUE #55: DJORNL parser: allow directional edges
The parser currently assumes that all edges are undirected, but the Jacobson team have revealed that in some cases, their data is directional. Fixing this in the parser and the DB will entail:

- Adding a new boolean field, `directed`, to the `djornl_edge` table to indicate whether an edge is directed or not; default will be undirected.
- Adding a new field to `csv_edge.yaml` to represent the column that will be added to the input files.
- Altering the logic in `store_parsed_edge_data` so that the edge IDs are only sorted if the edge is not directional.
- Adding tests for directional and undirected edges.

ialarmedalien:
Fixes #55.

[Commit 1](https://github.com/kbase/relation_engine/commit/3c8f0fa6fc4360e51e2edc1eb847b5d6d4310133): Adding black code formatter to dev requirements and running it prior to running the test suite.
Adding the 2020 resolver to the pip command as suggested by pip warning messages

[Commit 2](https://github.com/kbase/relation_engine/commit/e39f83a14608455d87e171f2669c4e0f01a41281): Add `directed` attribute to DJORNL edge spec, update the parser to be able to differentiate directed and undirected edges, and add in tests.

- [ ] I updated the README.md docs to reflect this change. - N/A

For changes to the codebase:

- [x] This is a breaking spec change and I have incremented the API version and added a summary to CHANGELOG.md.


importers/djornl/parser.py
```diff
@@ -312,11 +312,8 @@ def process_file(self, file, remap_fn, store_fn, err_list, validator=None):
             if validator is not None:
                 # validate the object
                 if not validator.is_valid(row_object):
-                    err_msg = "".join(
-                        f"{file['path']} line {line_no}: " + e.message
-                        for e in sorted(validator.iter_errors(row_object), key=str)
-                    )
-                    err_list.append(err_msg)
+                    for e in sorted(validator.iter_errors(row_object), key=str):
+                        err_list.append(f"{file['path']} line {line_no}: " + e.message)
```
ialarmedalien, line 316: slight reformatting of how errors are collected

importers/djornl/parser.py
```diff
@@ -364,6 +373,14 @@ def store_parsed_edge_data(self, datum):
             # report non-matching data
             return f"duplicate data for edge {edge_key}"
 
+        # create a unique key for the DB for this record
+        datum["_key"] = "__".join(
```
ialarmedalien, line 377: moving `_key` generation out of the remapping stage and into the storage stage to ensure that all values have already been transformed prior to creating the `_key`

spec/collections/djornl/djornl_edge.yaml
```diff
@@ -13,7 +13,7 @@ schema:
   title: Arabidopsis gene-gene or gene-phenotype edge
   description: Generic gene-to-gene or gene-to-phenotype edge for Dan Jacobson Arabidopsis data
   type: object
-  required: [score, edge_type, _from, _to, _key]
+  required: [score, edge_type, _from, _to, _key, directed]
```
ialarmedalien, line 16: edges must now have the `directed` field, which is either true or false (default: false)

spec/datasets/djornl/csv_edge.yaml
```diff
@@ -16,3 +16,10 @@ properties:
     pattern: ^\d+(\.\d+)?$
   edge_type:
     $ref: edge_type.yaml
+  directed:
```
ialarmedalien, line 19: in the CSV input file, the directed column contains either `1` or `0` (shorter than writing out `true` or `false`, plus no need to worry about case)

spec/datasets/djornl/definitions.yaml
```diff
@@ -21,20 +21,31 @@ definitions:
       pattern: ^(\S+__){3}(\S+)$
     _from:
       type: string
-      title: Gene ID
+      title: Node ID
+      format: regex
+      pattern: ^\S{2,}.*$
```
ialarmedalien, line 26: Adding in some extra constraints on IDs to ensure that there is some meaningful data in them

spec/datasets/djornl/definitions.yaml
```diff
@@ -21,20 +21,31 @@ definitions:
...
     _to:
       type: string
-      title: Gene or Phenotype ID
+      title: Node ID
+      format: regex
+      pattern: ^\S{2,}.*$
```
jayrbolton, line 31: Do these need to be single quoted so the backslashes aren't handled as escape chars? 
Assuming you tested this so I'm guessing not

jayrbolton: This was weirdly hard to find, but according to this page:
http://blogs.perl.org/users/tinita/2018/03/strings-in-yaml---to-quote-or-not-to-quote.html
In unquoted strings, backslashes do not need to be escaped

jayrbolton:
Hooray for the code formatter, but ideally we should add the formatter and all syntactic changes in a separate PR in the future, and add unrelated semantic changes to the code in another PR.

PR #61, opened 2020-12-15
Adding spec for sample_ontology_link collection in new samples folder.

slebras:

For changes to the codebase:

### DIFF
--- a/spec/collections/samples/sample_ontology_link.yaml
+++ b/spec/collections/samples/sample_ontology_link.yaml
@@ -1,17 +1,18 @@
 name: sample_ontology_link
 type: edge
+delta: true
 schema:
   "$schema": http://json-schema.org/draft-07/schema#
   type: object
-  required: [_from, _to, created]
+  required: [from, to, created, expired, sample_metadata_term]
   description: minimum necessary terms for sample -> ontology edge link.
   properties:
-    _from:
+    from:
       type: string
       description: the sample node _id (as found in sample_nodes collection)
       examples: ["samples_nodes/465b1476-3699-4e6c-a06b-8d384fcc41f3_6d5999ee-42fb-4bad-a3b9-901aa1b490c5_f4bc367798eb923f77d7405031723908"]
       pattern: ^[\w-]+\/[\w-]+_[\w-]+_[\w-]+
-    _to:
+    to:
       type: string
       description: The _id of an ontology vertex, such as from ENVO, GO, etc.
       examples: ["ENVO_terms/ENVO:01000221_v2019-03-14", 'GO_terms/GO:0047161_v2019-01-01']
--- a/spec/collections/samples/sample_ontology_link.yaml
+++ b/spec/collections/samples/sample_ontology_link.yaml
@@ -4,44 +4,58 @@ schema:
   "$schema": http://json-schema.org/draft-07/schema#
   type: object
   required: [_from, _to, created]
-  description: The _from sample node is associated with the _to taxon.
+  description: minimum necessary terms for sample -> ontology edge link.
   properties:
     _from:
       type: string
       description: the sample node _id (as found in sample_nodes collection)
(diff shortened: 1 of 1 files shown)


spec/collections/samples/sample_ontology_link.yaml
```diff
@@ -0,0 +1,47 @@
+name: sample_ontology_link
+type: edge
+schema:
+  "$schema": http://json-schema.org/draft-07/schema#
+  type: object
+  required: [_from, _to, created]
+  description: The _from sample node is associated with the _to taxon.
```
ialarmedalien, line 7: This would be better going somewhere else. It isn't necessary to have a description for every attribute in your schema. You can also consider putting in a `title`, `example`s, and validation for your attributes, if you want to do that.

Have a look at these schemas for examples:
- https://github.com/kbase/relation_engine/blob/develop/spec/collections/silva/silva_taxon.yaml
- https://github.com/kbase/relation_engine/blob/develop/spec/datasets/djornl/definitions.yaml

There's a great guide to JSON schema here if you want more info: https://json-schema.org/understanding-json-schema/

spec/collections/samples/sample_ontology_link.yaml
```diff
@@ -0,0 +1,47 @@
...
+  properties:
+    _from:
+      type: string
+      description: the sample node _id (as found in sample_nodes collection)
+    _to:
+      type: string
+      description: The _id of an ontology vertex, such as from ENVO, GO, etc.
+    created:
+      type: integer
+      description: timestamp of when the link was created
```
ialarmedalien, line 17: Timestamp in what format?

spec/collections/samples/sample_ontology_link.yaml
```diff
@@ -0,0 +1,47 @@
...
+  type: object
+  required: [_from, _to, created]
+  description: The _from sample node is associated with the _to taxon.
+  properties:
+    _from:
+      type: string
+      description: the sample node _id (as found in sample_nodes collection)
+    _to:
+      type: string
+      description: The _id of an ontology vertex, such as from ENVO, GO, etc.
```
ialarmedalien, line 14: Put in some examples here to make it clear.

spec/collections/samples/sample_ontology_link.yaml
jayrbolton: Probably should be edited to say "unix epoch" (i assume since it's an integer)

spec/collections/samples/sample_ontology_link.yaml
```diff
@@ -0,0 +1,47 @@
+name: sample_ontology_link
```
jayrbolton, line 1: Probably ask @scanon whether these should be "time travel" collections. If it's used in a narrative that should be (theoretically) reproducible, then probably they will want them to be time travel-able.

If Shane does want this, I'll tell you how to do the time travel style collections (it's not very different)

spec/collections/samples/sample_ontology_link.yaml
jayrbolton: I would also put "minimum: 0", cause why not

spec/collections/samples/sample_ontology_link.yaml
```diff
@@ -0,0 +1,47 @@
...
+      description: The _id of an ontology vertex, such as from ENVO, GO, etc.
+    created:
+      type: integer
+      description: timestamp of when the link was created
+    createdby:
+      type: string
+      description: Who made this sample-ontology link?
+    expired:
+      type: integer
+      description: timestamp of when this link expires
```
jayrbolton, line 23: Unix epoch, minimum 0

spec/collections/samples/sample_ontology_link.yaml
jayrbolton: I agree that you only need the property descriptions, but no biggie

jayrbolton:
Some minor comments, otherwise :+1:

PR #68, opened 2021-01-28
Additional collections for WITH clause to fix query

jayrbolton:
Fix the `ws_fetch_related_data` query by adding more collections in the WITH clause to appease cluster mode.

Tests do not cover this error because apparently running the DB as `arango --starter.local` does not require these explicit read locks. I don't have time to work on that right now, but we can add it to the queue.

Noticing pretty poor performance for this query in prod.

This is not a breaking API change

### DIFF
--- a/importers/djornl/parser.py
+++ b/importers/djornl/parser.py
@@ -802,7 +802,7 @@ Node data available
         value_width
     )
     if summary.get("errors_total") > 0:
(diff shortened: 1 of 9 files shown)


scanon (approved):
I fixed the error with the complaint about the type for the empty list.

PR #70, opened 2021-01-29
Show update source

jayrbolton:
Show the file path or URL of where we pulled the specs from when we run an update. This will be very useful for me, as often this can be a mystery and I have to debug the config.

We could write a test to show this works for a URL source, but I think not very necessary

For changes to the codebase:


ialarmedalien:
Can you remove the first two commits, since they are part of the other PR? Looks good otherwise.

jayrbolton:
This was branched from `simplify-workflow`. I'll unbranch it so it's more quickly mergeable

jayrbolton:
Done

jayrbolton:
Forgot the VERSION and CHANGELOG.md

PR #71, opened 2021-02-02
Add support for taxonomy and ontology data sources to address SCT-2918

eapearson:
This set of changes adds support for "data sources". Although data sources exist in the system in some measure, this adds support for a data sources collection, a pair of stored queries, an importer, and sample data.

Data sources correspond (1-1) with taxonomy and ontology namespaces. A data source describes the source, typically an organization, and may provide "additional fields" - fields which are additive to the base object.

Additionally a data source carries a "type" field, which has the value of either "taxonomy" or "ontology". Each data source type has an implicit (not defined here) base object. It is this base object to which the additional fields are considered to be added by the data source (namespace) implementation, and described by a data source document.

There is a companion update in the Taxonomy API which utilizes the stored query.

This set of changes also brings support for developing RE locally with a downstream consumer. This support is in the form of an alternative docker-compose file which utilizes the full set of collection specs, so it is ready to accept data for any collection. It is based on the testing docker-compose, so, like it, all service dependencies are self-contained. It also operates on the docker network "kbase-dev", which allows easy operation with services or ui components running on the same network.

I used this version of the RE server to load the data sources, and conduct integration testing with the Taxonomy Landing Page. That is something of a stand-in for an RE integration test covering the validity of the collection and dataset schemas.

The changes have been tested with Taxonomy API (in an upcoming PR), and with the Taxonomy Landing Page (via the Taxonomy API).

- [-] I updated the README.md docs to reflect this change.
  - I don't think this is relevant for this set of changes.

For changes to the codebase:

- [-] I have written tests to cover this change.
  - I don't think there are integration tests for individual collections and queries, are there?
  - The Taxonomy API has tests against the data sources stored queries, though.
- [-] This is a breaking API change and I have incremented the API version and added a summary to CHANGELOG.md.


ialarmedalien:
It seems to me that the UI-specific files (the docker-compose file, etc.) would be better off living in the UI repo -- e.g. see the organisation of https://github.com/kbaseapps/kb_cytoscape, which has a docker-compose override file that uses the compose set-up from this repo with an extra container and a few local modifications.

CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## NEXT
```
jayrbolton, line 8: Please change title to `## [Unreleased]`

importers/data_sources/importer.py
```diff
@@ -0,0 +1,153 @@
...
+
+        # The save_dataset method expects a list of documents
+        # to save, so we are already set!
+        schema_file = os.path.join(self.get_dataset_schema_dir(), "data_sources_nodes.yaml")
+        validator = get_schema_validator(schema_file=schema_file)
+
+        file_path = os.path.join(data_dir, 'data_sources.json')
+        with open(file_path, 'r') as data_file:
+            data_sources = json.load(data_file)
+            for data_source in data_sources:
```
jayrbolton, line 90: 90-94 can be unindented out of the context manager

importers/data_sources/importer.py
```diff
@@ -0,0 +1,153 @@
...
+                if not validator.is_valid(data_source):
+                    for e in sorted(validator.iter_errors(data_source), key=str):
```
jayrbolton, line 92: I think a more readable version would be to remove the for loop 

```py
err = sorted(validator.iter_errors(data_source), key=str)[0]
print(f'... {err.message}')
return
```
Unless you meant to print every error, in which case the `return` needs to be unindented

importers/data_sources/importer.py
```diff
@@ -0,0 +1,153 @@
...
+        return self._config
+
+    def get_dataset_schema_dir(self):
+        dir_path = os.path.dirname(os.path.realpath(__file__))
+        # TODO:  factor out "data_sources"
+        return os.path.join(
+            dir_path, "../", "../", "spec", "datasets", "data_sources"
+        )
+
+    def get_relative_dir(self, path):
```
jayrbolton, line 56: Pure-ish functions like this that make no use of `self` should either be class methods, or could be moved out of the class as a regular functions, so we can test or reuse them without needing an instance.

local-ui-dev/docker-compose.yaml
```diff
@@ -0,0 +1,65 @@
+version: "3"
+
+# This docker-compose is for developer convenience, not for running in production.
+networks:
```
jayrbolton, line 4: Can you just add some of these extra configuration lines, such as external network name and container name, to the root docker-compose.yaml and use that for local development? That root docker-compose is meant for development as well, and it looks like there is very little difference between this and that file.

local-ui-dev/README.md
```diff
@@ -0,0 +1,108 @@
...
+pip install -r dev-requirement.txt
+```
+
+Run your data importer. Each importer is an independent script, but they will probably follow a pattern similar to this example.
+
+This example uses the `data_sources.json` import data to populate the `data_sources_nodes` collection.
+
+Here we install from the default built-in data source, in dry-run mode, which prevents actual Arango changes:
+
+```bash
```
jayrbolton, line 44: It seems like there is a lot of redundant documentation between this file and the README found under data sources

jayrbolton (changes requested):
Review is still in progress. Overall things are looking good but I did have some comments. Will look more on Monday.

local-ui-dev/docker-compose.yaml
eapearson: Refactored this to use one docker-compose; enabled by adding external docker network creating as a make task and a dependency of all the docker-compose running tasks, and an env variable SPEC_RELEASE_PATH which can override the SPEC_RELEASE_PATH in docker-compose, since the local-ui-dev workflow relies on all collections being available.

importers/data_sources/importer.py
eapearson: That (and another method get_dataset_schema_dir) have since been moved out of the class.

importers/data_sources/importer.py
eapearson: I've refactored the error handling and message printing; it looks like I didn't quite finish up the error handling and left that return in while debugging!

importers/data_sources/README.md
```diff
@@ -0,0 +1,64 @@
+# Install data_sources dataset
+
+The script and data within this directory allow one to load the _data sources_ into the RE database. The default RE base url is `http://localhost:5000`, as defined in `importers/utils/config.py`, which can be redefined with the environment variable `RES_API_URL`. 
+
+This script has only been used in development thus far. In this case, the local RE Api container must be started via `make start-ui-dev`, as this starts the services with the `docker-compose.yaml` configuration located in `local-ui-dev`, which initializes all collections.
+
+Of course, in development you may have a different process; and in production I (Erik) have no clue how data is initialized or loaded. This script should work fine for initial load, and can be used for subsequent updates if the script is pointed at a data file which contains just the new documents.
```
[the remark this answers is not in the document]
eapearson: I must have removed the reference to that narrative, because it isn't in the file any longer. It was probably there as a aide-mémoire, but I don't remember now why.

importers/data_sources/importer.py
eapearson: The file access now handles OSError explicitly, and the dependent code has been moved out from the with block

spec/datasets/data_sources/data_sources_nodes.yaml
```diff
@@ -0,0 +1,2 @@
+$schema: http://json-schema.org/draft-07/schema#
+$ref: definitions.yaml#definitions/DataSource
```
ialarmedalien, line 2: please add in a line ending here.

scripts/run_tests.sh
```diff
@@ -6,10 +6,11 @@ set -e
 (cd /app/relation_engine_server/test/spec_release && \
   tar czvf spec.tar.gz sample_spec_release)
 
-black .
-flake8 --max-complexity 20 /app
-mypy --ignore-missing-imports /app
-bandit -r /app
+# TODO: why reformatting code?
+# black .
```
ialarmedalien, line 10: To ensure that the code has a uniform layout. It assumes that whoever is developing runs the tests regularly before committing code.

importers/data_sources/importer.py
```diff
@@ -0,0 +1,192 @@
+"""
+Loads data sources into RE.
+This is a simple importer, as the data source load data is quite simple, 
+and also in it's ultimate form, requiring no transformation.
```
ialarmedalien, line 4: typo: `s/it's/its/`

importers/data_sources/importer.py
```diff
@@ -0,0 +1,192 @@
...
+        if not resp.ok:
+            raise RuntimeError(resp.text)
+
+        note('success', f"Saved docs to collection {collection}!")
+        for key, value in resp.json().items():
+            note('info', f'     {key}: {value}')
+        return resp
+
+
+def main():
```
ialarmedalien, line 165: have you written any tests for module?

spec/datasets/data_sources/definitions.yaml
```diff
@@ -0,0 +1,163 @@
...
+        license,
+        item_link,
+        citation,
+        additional_fields,
+      ]
+    properties:
+      ns:
+        type: string
+        format: regex
+        pattern: "^[a-z0-9]+[a-z0-9_]*[a-z0-9]$"
```
ialarmedalien, line 80: will these always be lowercase?

spec/datasets/data_sources/definitions.yaml
```diff
@@ -0,0 +1,163 @@
...
+      A description of an external data source.
+      Data sources are external entities, usually governed and managed by an organization,
+      with data repositories, data policies, and specific data structures.
+      (e.g. taxonomy, ontology), individual data sources may provide data beyond this
+      general structure; the data source definition must provide a description
+      of those additional fields so that they can be utilized by interfaces.
+    type: object
+    additionalProperties: false
+    required:
+      [
```
ialarmedalien, line 63: switch this over to using the `- item` notation so that it is consistent with the other arrays in the file

spec/datasets/data_sources/definitions.yaml
```diff
@@ -0,0 +1,163 @@
+$schema: http://json-schema.org/draft-07/schema#
+name: definitions
+title: Data Sources schema definitions
+description: Node and edge metadata definitions for Data Sources
+definitions:
```
ialarmedalien, line 5: Nice!

spec/datasets/data_sources/definitions.yaml
```diff
@@ -0,0 +1,163 @@
...
+            required: [url, label]
+            properties:
+              url:
+                type: string
+                examples:
+                  - http://creativecommons.org/licenses/by-sa/4.0/
+              label:
+                type: string
+                examples:
+                  - Creative Commons Attribution-ShareAlike 4.0 International License
```
ialarmedalien, line 130: do these need to be outdented so they are on the same level as `description`? Same with the `item_link` below.

docs/local-ui-dev.md
```diff
@@ -0,0 +1,85 @@
...
+
+At present, only instructions for installing the `data_sources` are provided. Clearly a complete local workflow would involve installing major chunks of the RE graph.
+
+see [importers/data_sources/README.md](../importers/data_sources/README.md#procedure).
+
+## Point services to relation-engine-api
+
+Within the `kbase-dev` docker network, the RE api is associated with the host `relation-engine-api`, and is bound to port `5000` as KBase services typically are.
+
+### Example: Taxonomy RE API
```
ialarmedalien, line 31: Are there complementary docs in the taxonomy and kbase-ui repos? If not, I think it would be far more useful to have the docs there as people are more likely to be coming to the RE repo from those projects, rather than the other way around.

spec/collections/data_sources/data_sources_nodes.yaml
```diff
@@ -0,0 +1,60 @@
...
+    A description of an external data source.
+    Data sources are external entities, usually governed and managed by an organization,
+    with data repositories, data policies, and specific data structures.
+    (e.g. taxonomy, ontology), individual data sources may provide data beyond this
+    general structure; the data source definition must provide a description
+    of those additional fields so that they can be utilized by interfaces.
+  type: object
+  additionalProperties: false
+  required:
+    [
```
ialarmedalien, line 22: use the `- item` notation to be consistent

ialarmedalien (changes requested):
Please add some tests for the importer and for the queries.

spec/datasets/data_sources/definitions.yaml
eapearson: These were automatically formatted by some tool from the `-` style to the `[]` style, and it bugged me, but I decided not to touch them; now I've touched them!

spec/datasets/data_sources/definitions.yaml
eapearson: No, these are embedded in oneOf - the type is either null or the defined object. Unless I'm misunderstanding the question?

docs/local-ui-dev.md
eapearson: I find it more convenient to have instructions for different development scenarios in the service repo itself, but ymmv. 

It seems to me the service repo should have documentation describing different development scenarios. As a kbase-ui developer, I'm already aware that generally I can work ui and service components in an integrated manner - how easy that is depends on whether the service already provides that support (and that is not universal by any means.)
But it is a good point that the kbase-ui docs could have instructions for developing with specific services.

spec/collections/data_sources/data_sources_nodes.yaml
eapearson: Yes it was annoying to have the formatter add square brackets and then to have to manually change them back, but I did it :)

spec/datasets/data_sources/definitions.yaml
eapearson: All of the data source namespaces at present have been coerced to lower case, e.g.  GTDB -> gtdb, and I have no reason to expect that to change. The spec chooses a strictest implementation matching this fact; could always be loosened up later if there is a reason to use user case letters.

scripts/run_tests.sh
eapearson: Yes, I understand what it is doing, but I don't understand why one would want this over, for instanced, a developer choice to format-on save, or running black manually.  As a developer I would not expect a test to modify any code.
I noticed this as an issue because the scope is the entire repo, and if one is using a virtual environment it reformats all dependencies as well (and can get tripped up doing so.)

importers/data_sources/importer.py
eapearson: thanks, yes, my laptop keyboard is falling apart and doubles up spaces all the time when it's in a bad mood.

spec/collections/data_sources/data_sources_nodes.yaml
ialarmedalien: Thanks! 😁

docs/local-ui-dev.md
ialarmedalien: There are arguments for both sides -- I think the best solution is to make sure that docs are available in both repos. You're already very experienced with kbase-ui, so you will know all the ins and outs of it; other users (e.g. myself) might have set up kbase-ui locally but not know about additional services that can be enabled.

scripts/run_tests.sh
ialarmedalien: I don't think other users have employed a virtual environment, which is probably why it hasn't come up until now. The main advantage of having black and those various code format checks in the tests is that they are inescapable -- the tests will fail if they don't pass. A more proper place for them would be implemented in the dev's IDE, as git hooks, as some sort of auto-committed GHA, etc.

In lieu of that, can you alter the scope of the black command to limit it to py files from the RE repo, instead of every file in the known world?

spec/datasets/data_sources/definitions.yaml
ialarmedalien: No, it was just a question.

importers/data_sources/importer.py
eapearson: sorry, do you mean tests for importer.py? Yes, added today.

spec/datasets/data_sources/data_sources_nodes.yaml
eapearson: done

importers/data_sources/importer.py
ialarmedalien: Excellent, thanks -- I will re-review tomorrow.

scripts/run_tests.sh
ialarmedalien: If you add the command line param `--check` to the black command, it will just test the formatting, but not change it. That seems appropriate.

eapearson: Good idea, but Flake8 error messages are much better than black (it isn't a linter.)

As black compatibility seems to a goal of this repo, it would be better to use the flake8-black plugin which will do the linting and if that passes will see if black would trigger a reformat and report that as a familiar flake8 message.

I've added that change, and also the recommended flake8 settings for black compatibilit (https://black.readthedocs.io/en/stable/compatible_configs.html#flake8), which led to fixing a bunch of files which exceed the black default line length of 88. Most of those were comment lines, but a few were code lines.

ialarmedalien: These changes have made the PR rather large and confusing -- can you remove the last two commits from this PR (`git reset --soft HEAD~2` and then force push), and create a new PR with just the formatting stuff and module updates in it? I will merge that one and then you can rebase this PR off that. Thank you!

eapearson: Ok, I'll try to get to that this week.

scanon (approved):
This all seems fine.  I would just emphasize that (in the future), maybe do the linter changes as a separate PR since it really makes it hard to see the main changes.

importers/data_sources/importer.py
```diff
@@ -0,0 +1,209 @@
...
+def note(type, message):
+    icon = ''
+    if type == 'info':
+        icon = 'ℹ'
+    elif type == 'success':
+        icon = '✓'
+    elif type == 'warning':
+        icon = '⚠';
+    elif type == 'error':
+        icon = '🐛'
```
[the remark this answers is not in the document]
eapearson: What, no fun allowed in programming :)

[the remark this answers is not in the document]
eapearson: It is actually bore because it is more efficient - the message status indicators are single unicode characters.

importers/data_sources/importer.py
```diff
@@ -0,0 +1,209 @@
+"""
```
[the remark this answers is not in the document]
eapearson: I started with the only other existing importer in the codebase and stripped out all of the unnecessary stuff, adding back in what was needed. If it were written from scratch (I couldn't find docs for this), it may have been simpler.

tox.ini
```diff
@@ -1,2 +1,3 @@
 [flake8]
-max-line-length = 120
```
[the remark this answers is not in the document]
eapearson: I know, but the request was to restore black functionality, and this was required for black and flake8 to be compatible. Enabling black without having it rewrite files will generate errors, which can only be addressed by fixing the line length issue.
I would rather have left black disabled for the GHA build, but the request was to add it back in. The fact that the flake8 config was not black compliant was a mistake, afaik.
I'd say this is also a reason to try harder to have services written in the same language at KBase be compatible in terms of tooling and practices. Otherwise, it is hard when entering a project to discover these sorts of issues until late in the lifecycle of a set of changes (PR).

docs/local-ui-dev.md
eapearson: Consider from the other point of view.
There are not too many development scenarios for an individual service, and having them documented in the service can be very useful. I can't count the number of times I've consulted the documentation in the narrative related to local development in docker which I added many years ago, without complaint.
There are many services a user interface may interface with, and many of them have their own idiosyncratic tools, scripts, docker builds, dependencies.

scanon (approved):
LGTM

docs/local-ui-dev.md
ialarmedalien: @eapearson have you added docs to the appropriate repos? Can you point me to them?

importers/data_sources/test/test_data_sources_importer.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+class Test_data_sources_Importer(unittest.TestCase):
+    # @classmethod
+    # def setUp(cls):
+    #     # import the results file
+    #     # results_file = os.path.join(_TEST_DIR, "djornl", "results.json")
+    #     dir_path = os.path.dirname(os.path.realpath(__file__))
+    #     root_data_path = os.path.join(dir_path, "data")
+    #     with modified_environ(RES_ROOT_DATA_PATH=root_data_path):
```
ialarmedalien, line 79: can you remove this commented-out code?

importers/data_sources/test/test_data_sources_importer.py
```diff
@@ -0,0 +1,205 @@
...
+
+# tests:
+# happy path:
+# - load data
+
+
+
+
+# sad  path:
+# - load dta with all sorts of problems
```
ialarmedalien, line 205: rm comments

scripts/run_tests.sh
```diff
@@ -6,10 +6,11 @@ set -e
 (cd /app/relation_engine_server/test/spec_release && \
   tar czvf spec.tar.gz sample_spec_release)
 
-black .
-flake8 --max-complexity 20 /app
-mypy --ignore-missing-imports /app
-bandit -r /app
+# TODO: why reformatting code?
```
ialarmedalien, line 9: remove these comments

docs/local-ui-dev.md
```diff
@@ -0,0 +1,85 @@
+# Local UI Development
+
+This directory contains tools and instructions for using the RE api in a local development context which allows close operation with other kbase services in a custom docker network.
```
ialarmedalien, line 3: there are a couple of different ways of referring to the RE API in the docs -- can you standardise them all to be `RE API`?

importers/data_sources/test/test_data_sources_importer.py
```diff
@@ -0,0 +1,205 @@
...
+        with self.assertRaises(KeyError):
+            imp.get_config_or_fail('foo')
+        imp = make_importer('standard', warm_config=False)
+        self.assertIsNotNone(imp.get_config_or_fail('API_URL'))
+
+    def test_happy_load(self):
+        imp = make_importer('standard')
+        self.assertTrue(imp)
+        result = imp.load_data(dry_run=True)
+        self.assertTrue(result)
```
ialarmedalien, line 100: can you test the imported data a bit more thoroughly?

importers/data_sources/test/test_data_sources_importer.py
```diff
@@ -0,0 +1,205 @@
...
+        imp = make_importer('bad_dir')
+        result = imp.load_data(dry_run=True)
+        self.assertFalse(result)
+
+    def test_bad_ns(self):
+        imp = make_importer('bad_ns')
+        result = imp.load_data(dry_run=True)
+        self.assertFalse(result)
+        result = imp.load_data(dry_run=False)
+        self.assertFalse(result)
```
ialarmedalien, line 118: Ditto with these tests -- can you add in something that ensures that what you want to be tested is _actually_ being tested?

importers/data_sources/test/test_data_sources_importer.py
```diff
@@ -0,0 +1,205 @@
...
+
+    @responses.activate
+    def test_do_import_dry_run(self):
+        responses.add(responses.PUT, f'{API_URL}/api/v1/documents',
+                      json={'do_not': 'care'}, status=200)
+        do_import(dry_run=False)
+        self.assertTrue(True)
+
+    @responses.activate
+    def test_do_import_dry_run(self):
```
ialarmedalien, line 148: I'm a bit surprised python doesn't care that this function has the same name as the two functions above!

importers/data_sources/importer.py
```diff
@@ -0,0 +1,209 @@
...
+Loads data sources into RE.
+
+This is a simple importer, as the data_sources load data is quite simple,
+and also very similar to its ultimate form, requiring little transformation.
+
```
ialarmedalien, line 6: can you write some brief usage docs for this module?

importers/data_sources/test/test_data_sources_importer.py
```diff
@@ -0,0 +1,205 @@
...
+        result = imp.load_data(dry_run=False)
+        self.assertFalse(result)
+
+    @responses.activate
+    def test_save_docs(self):
+        responses.add(responses.PUT, f'{API_URL}/api/v1/documents',
+                      json={'do_not': 'care'}, status=200)
+        imp = make_importer('standard')
+        result = imp.load_data(dry_run=False)
+        self.assertTrue(result)
```
ialarmedalien, line 126: can you add in a check somewhere that ensures that data being loaded is as you expect it to be?

spec/stored_queries/data_sources/data_sources_get_data_sources.yaml
```diff
@@ -0,0 +1,22 @@
+# Get 0 or more data sources matching the given type, optionally
+# filtered by an array of namespace ids
+name: data_sources_get_data_sources
```
ialarmedalien, line 3: Can you write a couple of quick tests for these queries?

Nice example tests: https://github.com/kbase/relation_engine/blob/develop/spec/test/stored_queries/test_ontology.py

importers/data_sources/importer.py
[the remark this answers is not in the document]
ialarmedalien: There aren't really any standardised importers -- it's all been somewhat ad hoc.

ialarmedalien:
When you have made the changes, please run black over the files that you have changed so that they are internally consistent, formatting-wise. Thanks!

scripts/run_tests.sh
eapearson: unfortunately, reverting the prior to commits also reverted some changes like this.

importers/data_sources/importer.py
eapearson: Hmm, there is a README.md, is that not sufficient? (I am tweaking it a bit too.)

spec/stored_queries/data_sources/data_sources_get_data_sources.yaml
eapearson: There are integration tests now, which exercises the importer and check the state of the db afterwards with re api and direct Arango calls. The queries are in those code pathways.

importers/data_sources/test/test_data_sources_importer.py
eapearson: Well, by definition any data successfully added would be compliant with the schema, but I get your point. 
There are now integration tests which validate not only successful addition of data, but also ensure that the results of the queries are compliant with the same schemas (since there is no transformation in the query, just the straight document.)

importers/data_sources/test/test_data_sources_importer.py
eapearson: have been renamed.

docs/local-ui-dev.md
eapearson: done

scripts/run_tests.sh
```diff
@@ -6,18 +6,27 @@ set -e
...
 coverage erase
+
 # spec validation
 python -m spec.validate
+
 # run importer/, relation_engine_server/, and spec/ tests
-coverage run --branch -m unittest discover -v
+coverage run --branch --parallel-mode -m unittest discover -v importers
+coverage run --branch --parallel-mode -m unittest discover -v relation_engine_server
+coverage run --branch --parallel-mode -m unittest discover -v spec
```
ialarmedalien, line 26: why did this change?

test/integration/utils/clock.py
```diff
@@ -0,0 +1,10 @@
+class Clock:
+    def __init__(self):
+        self.clocks = ["🕛", "🕐", "🕑", "🕒", "🕓", "🕔", "🕕", "🕖", "🕗", "🕘", "🕙", "🕚"]
```
ialarmedalien, line 3: Hope Shane doesn't see these!

test/integration/utils/wait_for_re_api.py
```diff
@@ -0,0 +1,47 @@
...
+RE_API_URL = os.environ.get("RE_API_URL")
+if RE_API_URL is None:
+    print('the "RE_API_URL" environment variable is required')
+    sys.exit(1)
+
+
+def wait_for_re_api() -> None:
+    """wait for a service or list of services to start up"""
+    TIMEOUT
+    POLL_INTERVAL
```
ialarmedalien, line 23: ?? what are these here for?

common/json_validation.py
```diff
@@ -0,0 +1,177 @@
+"""
```
ialarmedalien, line 1: This duplicates what is in `relation_engine_server/utils/json_validation.py` -- correct? Did you move that file to here and git doesn't realise what's happened, or is this a copy?

importers/data_sources/importer.py
```diff
@@ -0,0 +1,272 @@
...
+- spec/datasets/data_sources/data_sources_nodes.yaml - definition of each data_source
+      being loaded as a DataSource object
+- spec/collections/data_sources/data_sources_nodes.yaml - replication of the DataSource
+      object with the addition of a  _key property; all fields by reference to the
+      definitions in spec/datasets/data_sources
+
+TODO: A DRYer design would be for the schemas in spec/collections to use an allOf to
+extend the base DataSource type in spec/datasets.
+However, the schema for collection definition files located in spec/collections
+expects that the schemas in spec/collections be be an object.
```
ialarmedalien, line 19: `be be` -- repeated word

scripts/run_tests.sh
eapearson: I don't remember if this was to specifically to avoid tests picking up a venv directory or the test directory, but generally to constrain the tests to the specified directories without any surprises; they are separate invocations because discover allows a single root directory (I'm pretty sure I would have tried that first!)

ialarmedalien: Cool, thanks for the explanation!

importers/data_sources/importer.py
eapearson: Gone now, but interestingly not an uncommon word duplication - can be found in the black and git libraries :)

docs/testing.md
```diff
@@ -0,0 +1,30 @@
...
+
+### Quick 
+
+Ensure dependencies:
+- make
+- python 3.7 or higher (?)
+
+```bash
+python -m venv venv
+source vevn/bin/activate
```
ialarmedalien, line 13: is it necessary to have a `venv` running now?

importers/data_sources/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Procedure
+
+Set up a Python environment:
+
+At top level of repo:
+
+```bash
+python -m venv venv
+source venv/bin/activate
```
ialarmedalien, line 19: there's no way of running the importers off the docker image so that you don't have to set up a virtual environment locally?

importers/test/test_djornl_parser.py
```diff
@@ -72,7 +72,7 @@ def test_config(self):
             parser.config("bananas")
 
     def test_load_no_manifest(self):
-        """test loading when the manifest does not exist"""
```
ialarmedalien, line 75: ???

can you restore this comment? Looks like a merge mistake.

scripts/run_tests.sh
```diff
@@ -6,18 +6,31 @@ set -e
...
-black .
-flake8 --max-complexity 20 /app
-mypy --ignore-missing-imports /app
-bandit -r /app
+# Run code quality checks before tests; a failure here
+# will prevent the tests from running.
+source_dirs="/app/relation_engine_server /app/client_src /app/spec"
+flake8 --max-complexity 20 $source_dirs
+mypy --ignore-missing-imports $source_dirs
+bandit -r $source_dirs
```
ialarmedalien, line 14: can you add in the black code formatting check here?

test/integration/README.md
```diff
@@ -0,0 +1,52 @@
+# Integration Tests
```
ialarmedalien, line 1: This looks great.

test/integration/tests/test_data_sources_import.py
```diff
@@ -0,0 +1,295 @@
...
+
+from relation_engine_server.utils.json_validation import (
+    get_schema_validator,
+)
+
+#
+# We are already running Python, of course, but we need to invoke another
+# python instance in a subprocess.
+#
+PYTHON_BIN = os.environ.get("PYTHON_BIN", sys.executable)
```
ialarmedalien, line 18: no existence check for this env var?!

test/integration/utils/wait_for_re_api.py
```diff
@@ -0,0 +1,45 @@
+"""
+Block until all dependent services come online.
```
ialarmedalien, line 2: is there no way to integrate this with the existing `wait_for` code?

docs/testing.md
eapearson: nope!

ialarmedalien: Excellent! Can you add something to this document so it's clear that you don't have to have the python stuff installed locally, and it can be run just using the docker image?

test/integration/utils/wait_for_re_api.py
eapearson: Yes, of course; I broke it out because I needed changes (e.g. utilize RE_API_URL), and the original one handles more cases than integration tests need.

test/integration/tests/test_data_sources_import.py
eapearson: It uses the default of sys.executable?

scripts/run_tests.sh
eapearson: done, also added as make task for running manually w/o having to invoke tests

importers/test/test_djornl_parser.py
eapearson: done.

importers/data_sources/README.md
eapearson: sure, just time and effort. yup, now down.

scripts/run_tests.sh
eapearson: I do think it would be best if black during tests would run in read-only mode and then have the dev do black formatting manually. I did not see a black option to do so. Of course, one could do something funky like copy the source code, run black, and inspect for modified files.

docs/testing.md
eapearson: Yes, the testing doc is rather skeletal, but the python manual testing instructions are gone.

scripts/run_tests.sh
ialarmedalien: I'm sure there is one as I remember mentioning it to you... let me see if I can find it.

ialarmedalien: yes: https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#exit-code

test/integration/utils/wait_for_re_api.py
ialarmedalien: It'd be far preferable not to have functionality duplicated in different places in the codebase -- it's more code to maintain, more potential for bit rot and cruft to build up, etc.

ialarmedalien:
@eapearson  it looks like there are only two outstanding issues in this PR -- this [small edit](https://github.com/kbase/relation_engine/pull/71#discussion_r690526100) and some [consolidation of functionality](https://github.com/kbase/relation_engine/pull/71#discussion_r688594120) here.

PR #83, opened 2021-05-12
adding script for auto-generating yaml files for ontology loading

zhlu9890:

For changes to the codebase:

- [X] This is not a breaking API change OR


ialarmedalien (approved):
LGTM

scanon:
This is a good start.  I think we need a few things.

* Some unit tests for the generation script.  It could start simple and we could build out.
* Documentation.  This should clarify the steps to add an ontology.
* I think the script could use some additional documentation too.

scripts/prepare_ontology.py
```diff
@@ -0,0 +1,81 @@
...
+import os
+import yaml
+import json
+import shutil
+
+"""
+python3 scripts/prepare_ontology.py scripts/test/data/data_sources.json fake_ontology
+"""
+
+__NAME = "__NAME__"
```
ialarmedalien, line 11: This is slightly confusing -- can you rename the variable something like `PLACEHOLDER` instead?

I would also not use both all caps *and* double underscores for variable names -- all caps is sufficient to differentiate them from other variables in this file.

scripts/prepare_ontology.py
```diff
@@ -0,0 +1,81 @@
...
+
+def clean_up_data(path):
+    if os.path.exists(path):
+        if os.path.isfile(path):
+            os.remove(path)
+        elif os.path.isdir(path):
+            shutil.rmtree(path)
+
+
+if __name__ == "__main__":
```
ialarmedalien, line 80: Can you put in some error handling?

scripts/test/test_prepare_ontology.py
```diff
@@ -0,0 +1,46 @@
...
+    parse_input,
+    parse_namespace,
+    clean_up_data,
+)
+
+_TEST_DIR = "/app/scripts/test"
+_TEST_NAMESPACE = "fake_ontology"
+
+
+class Test_prepare_ontology(unittest.TestCase):
```
ialarmedalien, line 20: can you add in some non-happy-path tests to ensure that any errors are caught and handled correctly by the code?

scripts/prepare_ontology.py
scanon: Can you clarify where you want the handling?  Since this would be run interactively, it may be better for the script to throw the error so any issues could get diagnosed more easily.

zhlu9890:
I added some tests and update the scripts as suggested. As Shane said, it would be running interactively in command line, so I didn't add much error handling but added warning and error msg to help user to find potential problems (most likely due to path privilege which user could fix accordingly).

ialarmedalien (approved):
Looks great! Merge away. 😄

PR #86, opened 2021-08-05
Update edge_type.yaml to align with exascale_data repo

izaakm:
- [ ] I updated the README.md docs to reflect this change.

For changes to the codebase:

- [ ] I have written tests to cover this change.
- [x] This is not a breaking API change OR
- [ ] This is a breaking API change and I have incremented the API version and added a summary to CHANGELOG.md.

Relation engine:

```
 $ RES_ROOT_DATA_PATH=../exascale_data/prerelease/ python -m importers.djornl.parser --dry-run
Parsing edge file /Users/m8z/src/exascale_data/prerelease/edge_data/AT-UU-KS-00-AA-01.tsv
Parsing edge file /Users/m8z/src/exascale_data/prerelease/edge_data/AT-UU-GA-01-AA-01.tsv
Parsing edge file /Users/m8z/src/exascale_data/prerelease/edge_data/AT-UU-PP-00-AA-01.tsv
Parsing edge file /Users/m8z/src/exascale_data/prerelease/edge_data/AT-UU-RE-00-AA-01.tsv
Parsing edge file /Users/m8z/src/exascale_data/prerelease/edge_data/AT-UU-RP-03-AA-01.tsv
Parsing edge file /Users/m8z/src/exascale_data/prerelease/edge_data/AT-UU-CD-00-AA-01.tsv
Parsing edge file /Users/m8z/src/exascale_data/prerelease/edge_data/AT-UU-RX-00-AA-01.tsv
Parsing node file /Users/m8z/src/exascale_data/prerelease/aranet2-aragwas-MERGED-AMW-v2_091319_nodeTable.csv
Parsing cluster file /Users/m8z/src/exascale_data/prerelease/cluster_data/out.aranetv2_subnet_AT-CX_top10percent_anno_AF_082919.abc.I2_named.tsv
Parsing cluster file /Users/m8z/src/exascale_data/prerelease/cluster_data/out.aranetv2_subnet_AT-CX_top10percent_anno_AF_082919.abc.I4_named.tsv
Parsing cluster file /Users/m8z/src/exascale_data/prerelease/cluster_data/out.aranetv2_subnet_AT-CX_top10percent_anno_AF_082919.abc.I6_named.tsv

 26745 Total nodes
711040 Total edges
 25108 Nodes in edge
---
Node Types
  5069 No type
 21669 gene
     7 pheno
---
Edge Types
 94952 AT-UU-KS-00-AA-01
 82900 AT-UU-GA-01-AA-01
[...]

PR #87, opened 2021-08-10
2021 08 updates

dakotablair:
The node data is being loaded in a new format, and this PR updates the importer and tests to support this. It also includes the latest edge type metadata to support the networks added in early August.


dakotablair:
This PR includes the changes from PR #86.

PR #88, opened 2022-01-12
WIP

n1mus:
WIP

WIP - forgot run_tests_simple script

WIP - instance None fixed

WIP - add analyzer

WIP - fulltext tests passing

For changes to the codebase:

- [ ] This is not a breaking API change OR


.github/workflows/run_tests.yaml
```diff
@@ -1,49 +0,0 @@
-name: Relation Engine test and deploy
```
ialarmedalien, line 1: why delete this file?

relation_engine_server/utils/json_validation.py
```diff
@@ -73,9 +68,9 @@ def get_schema_validator(schema=None, schema_file=None, validate_at=""):
     validation_schema = resolve_pointer(schema, validate_at)
 
     if schema_file:
-        resolver = ExtendedRefResolver(schema_file, schema)
+        resolver = RefResolver(schema_file, schema)
     else:
-        resolver = ExtendedRefResolver.from_schema(schema)
+        resolver = RefResolver.from_schema(schema)
```
ialarmedalien, line 73: does the updated version of the `jsonschema` lib now parse `yaml` files as well as `json`?

spec/collections/gtdb/gtdb_taxon.yaml
```diff
@@ -22,11 +26,21 @@ schema:
...
+    name:
       type: string
       description: The name of the taxon
       examples: ['Firmicutes', 'Sediminibacterium sp002786355']
-    rank:
+    first_version:
       type: string
-      title: Taxonomic rank
-      examples: [domain, phylum]
+      examples: [r89]
```
ialarmedalien, line 39: could you put in a description for this and `last_version`?

spec/test/stored_queries/query_dev/test_fulltext_search_dev.py
```diff
@@ -0,0 +1,486 @@
+import os
```
ialarmedalien, line 1: is this file needed or just an accidental commit?

Dockerfile
```diff
@@ -24,6 +24,7 @@ RUN apk --update add --virtual build-dependencies build-base python3-dev && \
     if [ "$DEVELOPMENT" ]; then pip install --no-cache-dir -r /tmp/dev-requirements.txt; fi && \
     apk del build-dependencies && \
     pip install types-requests types-PyYAML
+RUN pip install pytest
```
ialarmedalien, line 27: is there any advantage to using pytest vs the standard `unittest` test harness?

relation_engine_server/utils/arango_client.py
```diff
@@ -100,27 +100,35 @@ def create_collection(name, config):
...
+def _create_indexes(coll_name, config):
+    """Create indexes for a collection"""
+    url = _CONF["api_url"] + "/index"
+    indexes = _get_indexes(coll_name)
     for idx_conf in config["indexes"]:
-        if _index_exists(idx_conf, indexes):
-            continue
         idx_type = idx_conf["type"]
         idx_url = url + "#" + idx_type
-        idx_conf["type"] = idx_type
```
ialarmedalien, line 117: 😖

relation_engine_server/utils/arango_client.py
```diff
@@ -100,27 +100,35 @@ def create_collection(name, config):
...
-    url = _CONF["api_url"] + "/index"
-    # Fetch existing indexes
-    auth = (_CONF["db_user"], _CONF["db_pass"])
-    resp = requests.get(url, params={"collection": coll_name}, auth=auth)
+def _get_indexes(coll_name):
+    """Fetch existing indexes for a collection"""
+    resp = requests.get(
+        url=_CONF["api_url"] + "/index",
+        params={"collection": coll_name},
+        auth=(_CONF["db_user"], _CONF["db_pass"]),
```
ialarmedalien, line 108: if this is used numerous times in this file, it might be worth turning it into a variable

spec/test/helpers.py
```diff
@@ -78,7 +78,8 @@ def check_spec_test_env():
         # copy the contents of /app/spec into /spec
         shutil.rmtree(_CONF["spec_paths"]["root"], ignore_errors=True)
         shutil.copytree("/app/spec", _CONF["spec_paths"]["root"])
-        download_specs()
+        if do_download_specs:
+            download_specs()
```
ialarmedalien, line 82: are there any cases where `do_download_specs` is not `True`?

spec/test/stored_queries/test_fulltext_search.py
```diff
@@ -0,0 +1,226 @@
+"""g
```
ialarmedalien, line 1: `g`?

.github/workflows/run_tests.yaml
n1mus: Because I wanted to lug it around local/docker03 without pushing to kbase/relation_engine because that tells everyone it failed and I couldn't push it to n1mus/relation_engine with the workflow because my token wouldn't allow it. Is temporary

relation_engine_server/utils/json_validation.py
n1mus: I don't think so ... I only updated `jsonschema` trying to fix that obtuse bug where the `jsonschema` error message was not helpful. In order to run things though, I had to get rid of some no-longer-existing `from jsonschema.compat import` things that were being used in `ExtendedRefResolver`, which I found out wasn't being used anywhere, serendipitously. So I axed `ExtendedRefResolver`. I could downgrade `jsonschema` again now that the bug was figured out, but I figure why not just keep things updated ...

spec/test/stored_queries/query_dev/test_fulltext_search_dev.py
n1mus: This file represents weeks of work I put into making the `generic/fulltext_search.yaml` stored query. I was planning on hooking it up with `make`. Its purpose is to connect with the `ci` database and test the stored query against every single current (as of time of coding) species/strain name in the `ncbi_taxon` collection; collect stats; stratify by complicated/simple scinames and all their prefixes and all single alphanumeric characters; and output mean/median execution times. There's also functionality for graphing the execution times depending on certain token criteria.

Dockerfile
n1mus: only if you think the xunit methodology is musty

n1mus: Oh wait, yeah, I like pytest because it gives better stack traces. There's probably other reasons too I'm not recalling

relation_engine_server/utils/arango_client.py
n1mus: This was originally because I was thinking of adding a check on ArangoDB server startup to check that all the indexes in the specs are existing as configured, since the indexes/analyzers/views are, in live deploy, actually manually added by an admin. I actually have no idea how I would actually do this in tests though because the indexes/analyzers/views are added in `setUpClass` so I might abandon this effort unless someone objects or can help.

spec/test/helpers.py
n1mus: Umm this was left in by accident probably

spec/test/stored_queries/test_fulltext_search.py
n1mus: No idea ... maybe I was in edit mode in vim when I thought I was in navigation mode

spec/test/stored_queries/query_dev/test_fulltext_search_dev.py
ialarmedalien: ok, great. I didn't read the contents due to the file name, but it sounds like it just needs renaming. 👍

relation_engine_server/utils/json_validation.py
ialarmedalien: We should check that everything still works correctly without the ExtendedRefResolver -- it was put in place because the jsonschema package only parses JSON files, and a number of our conf files are in yaml (as you will know!). IIRC the ExtendedRefResolver is a fairly simple override of the existing jsonschema RefResolver with two additions: 1) assume that if there's no scheme for a ref, it's a local file, and 2) if the file name ends with '.yml' or '.yaml', use the yaml parser to open it.

See https://python-jsonschema.readthedocs.io/en/latest/_modules/jsonschema/validators/#RefResolver.resolve_remote

spec/test/stored_queries/test_fulltext_search.py
```diff
@@ -0,0 +1,226 @@
...
+        params={"stored_query": "fulltext_search"},
+        data=json.dumps(data),
+    )
+
+    if "error" in resp.json():
+        raise ArangoServerError(resp.text)
+
+    docs = resp.json()['results']
+    hits = [doc[search_attrkey] for doc in docs]
+    self.assertIn(search_text, hits)
```
ialarmedalien, line 225: should you check that the search text is in each doc in the `hits` array?

ialarmedalien: I think ideally you'd specify which results you expect the search to turn up, so you wouldn't just check that the search text is in each doc, but rather that you have all the docs returned that you would expect to see.

spec/test/stored_queries/test_fulltext_search.py
```diff
@@ -0,0 +1,226 @@
...
+                ts=_NOW if sciname in scinames_test_latest else None,
+                filter_attr_expr=[{'rank': 'species'}, {'rank': 'strain'}, {'strain': True}],
+                offset=0,
+                limit=LIMIT,
+                select=['id', 'scientific_name']
+            )
+
+    def test_stored_query_validation_fail(self):
+        for sciname in scinames_test_all:
+            with self.assertRaises(Exception) as e:
```
ialarmedalien, line 141: is there specific message text you can check for?

ialarmedalien:
Is this PR still needed? If not, can you close it and delete the branch?

PR #89, opened 2022-01-13
Fulltext search stored query + icu_tokenize analyzer

n1mus:

For changes to the codebase:

- [n/a] This is a breaking API change and I have incremented the API version and added a summary to CHANGELOG.md.

### DIFF
--- a/relation_engine_server/test/spec_release/sample_spec_release/spec/collections/ncbi/ncbi_taxon.yaml
+++ b/relation_engine_server/test/spec_release/sample_spec_release/spec/collections/ncbi/ncbi_taxon.yaml
@@ -5,6 +5,7 @@ delta: true
 indexes:
   - type: fulltext
     fields: [scientific_name]
+    minLength: 1
   - type: persistent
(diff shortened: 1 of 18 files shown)


requirements.txt
```diff
@@ -3,6 +3,7 @@ greenlet==0.4.16
 gunicorn==19.9.0
 gevent==1.3.7
 simplejson==3.16.0
+pytest==6.2.5
```
n1mus, line 6: I don't think this is used but I have used it in temporary scripts

relation_engine_server/test/spec_release/sample_spec_release/spec/collections/ncbi/ncbi_taxon.yaml
```diff
@@ -5,6 +5,7 @@ delta: true
 indexes:
   - type: fulltext
     fields: [scientific_name]
+    minLength: 1
```
n1mus, line 8: All the fulltext indexes can get a new param `minLength=1` in order to work well in searches

PR #97, opened 2022-02-08
recursive obj subsetting

n1mus:

For changes to the codebase:

### DIFF
--- a/relation_engine_server/main.py
+++ b/relation_engine_server/main.py
@@ -42,7 +42,7 @@ def return_error(error_dict, code):
 
 @app.route("/", methods=["GET"])
 def root():
-    """Server status."""
+    """Server status. develop is default branch"""
     if os.path.exists(".git/refs/heads/develop"):
         with open(".git/refs/heads/develop", "r") as fd:
             commit_hash = fd.read().strip()
--- a/relation_engine_server/main.py
+++ b/relation_engine_server/main.py
@@ -43,8 +43,8 @@ def return_error(error_dict, code):
 @app.route("/", methods=["GET"])
 def root():
     """Server status."""
-    if os.path.exists(".git/refs/heads/master"):
-        with open(".git/refs/heads/master", "r") as fd:
+    if os.path.exists(".git/refs/heads/develop"):
+        with open(".git/refs/heads/develop", "r") as fd:
             commit_hash = fd.read().strip()
     else:
         commit_hash = "unknown"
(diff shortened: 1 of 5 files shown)


relation_engine_server/main.py
```diff
@@ -43,8 +43,8 @@ def return_error(error_dict, code):
 @app.route("/", methods=["GET"])
 def root():
     """Server status."""
-    if os.path.exists(".git/refs/heads/master"):
-        with open(".git/refs/heads/master", "r") as fd:
+    if os.path.exists(".git/refs/heads/develop"):
```
n1mus, line 46: So you can get the right commit version when `curl`ing instead of just getting "unknown" commit. Note to self: document this in later PR

relation_engine_server/utils/config.py
```diff
@@ -31,7 +31,6 @@ def get_config():
     db_readonly_user = os.environ.get("DB_READONLY_USER", db_user)
     db_readonly_pass = os.environ.get("DB_READONLY_PASS", db_pass)
     api_url = db_url + "/_db/" + db_name + "/_api"
-    print(db_user, db_pass)
```
n1mus, line 34: Was being spewed into logs

relation_engine_server/utils/wait_for.py
```diff
@@ -41,7 +41,6 @@ def wait_for_service(service_list: List[str]) -> None:
             try:
                 conf = service_conf[name]
                 auth = (_CONF["db_user"], _CONF["db_pass"])
-                print("auth is", auth)
```
n1mus, line 44: Was showing up in logs

relation_engine_server/utils/ensure_specs.py
```diff
@@ -228,6 +228,25 @@ def excise_namespace(analyzer_name: str) -> str:
...
+    if isinstance(l, dict) and isinstance(r, dict):
+        return all(
+            [k in r.keys() and is_obj_subset_rec(l[k], r[k]) for k in l.keys()]
+        )  # ignore: typing
+    elif isinstance(l, list) and isinstance(r, list):
+        return all([le in r for le in l])
+    else:
+        return l == r  # noqa: E741
+
+
```
n1mus, line 249: My first guess on how to compare the local/server specs was just `local.items() <= server.items()` with the assumption that ArangoDB would only add default parameters. Kind of worked, but ended up working better with recursive subset checks ..

relation_engine_server/main.py
ialarmedalien: Can you just add a comment in the code here?

ialarmedalien (changes requested):
Just add a comment about checking the `develop` branch rather than `master`

PR #101, opened 2022-03-02
prep release and DATAUP-706-doc-startup

n1mus:

For changes to the codebase:

- [x] This is a breaking API change and I have incremented the API version and added a summary to CHANGELOG.md.


README.md
```diff
@@ -17,6 +17,14 @@ These specifications are used by the [Relation Engine API](relation_engine_serve
 
 The relation engine server (`relation_engine_server/`) is a simple API that allows KBase community developers to interact with the Relation Engine graph database. You can run stored queries or do bulk updates on documents.
 
+## Relation Engine Startup
+* Docker image is built with environment variable `SPEC_RELEASE_PATH=/opt/spec.tar.gz'. This contains the specs from the repo itself.
+* Wait for response from auth, workspace, and arangodb services, as they are set up
+* Specs are set up. Either the repo specs or remote specs are loaded into the specs root path
+* Collections, views, and analyzers from the specs are added to the ArangoDB server. If the collection, view, or analyzer already exists, but in a different configuration, it will _not_ be overwritten.
+* Collections, views, and analyzers from the loaded specs are compared to those on the ArangoDB server. If the loaded specs' attributes are not recursively a subset of the server specs, then an exception is raise. (This is just preliminary validation behavior.)
```
ialarmedalien, line 25: an exception is **raised**

I'd incorporate the last sentence into the previous two, so you have something like:

> Basic validation of the collections, views, and analyzers is performed by comparing the loaded specs to those on the ArangoDB server. If the loaded specs' attributes are not recursively a subset of the server specs, an exception is raised and ... (server start up fails? what happens if there's an exception?)

n1mus: Ok ... that sounds good ... but I'm gonna call perfect is the opposite of good

PR #102, opened 2022-03-17
Long query testing -- violin plots comparing old and new query performance

n1mus:

For changes to the codebase:

Not sure if this belongs in the main repo. It's just refactoring of the long query tests and then adding graphing of results.


spec/test/stored_queries/test_query.py
```diff
@@ -102,7 +118,10 @@
 
 
 def use_sort(search_text):
-    """Determine whether to use the sorting or non-sorting query"""
+    """
+    Determine whether to use the sorting or non-sorting stored query for the new query.
+    Smaller search texts' results will not be sorted on.
```
ialarmedalien, line 123: There's a handy VSCode extension that I've been using to do this - "autoDocstring: VSCode Python Docstring Generator"

PR #103, opened 2022-03-18
Special Chars for Fulltext Search

n1mus:

For changes to the codebase:

The FULLTEXT search has some special escape characters, e.g., `,`, `+`, `-`, `|`, `:`, `"`, `'`. Add some tests for any special character that has not already been tested and exists in a `ncbi_taxon` sciname. (Note this is a bare minimum test, because the search could behave differently within a vaster collection of scinames, and depending on the syntax context of the special escape char.)


spec/test/stored_queries/test_fulltext_search.py
```diff
@@ -55,6 +56,17 @@
     "Vaccinia virus WR 65-16",
     "Dengue virus 2 Jamaica/1409/1983",
     "Dengue virus 2 Thailand/NGS-C/1944",
+    # --- Escape chars (,:+-|"') ---
```
ialarmedalien, line 59: are `(` and `)` also escape chars? Could you put them together and/or space-separate all the characters to be escaped, so it's clear that you're not using them as parentheses?

spec/test/stored_queries/test_fulltext_search.py
```diff
@@ -55,6 +56,17 @@
...
+    # --- TODO sample scinames with the escape chars in different variety of syntaxes ---
+    "Salmonella enterica subsp. diarizonae serovar 60:r:e,n,x,z15",
+    "Fusarium cf. solani 3+4-uuu DPGS-2011",
+    "Integrating expression vector pJEB403+drrA",
+    "Vector pEntry-attR2-IRES-eGFP-luc+-pA-attL3",
+    "low G+C Gram-positive bacterium HTA462",
+    "Reporter vector p1168hIL6mC/EBP-luc+",
+    "Pleurocapsales cyanobacterium 'Beach rock 4+5\"'",
+    "Nostoc sp. 'Peltigera sp. \"hawaiensis\" P1236 cyanobiont'",
+    "|Fake|fake|fake|  ||fake||",
```
ialarmedalien, line 69: there's an extra `i` in `technically` on the next line

PR #108, opened 2022-03-29
Update dev-requirements.txt

bio-boris:
See https://github.com/psf/black/issues/2964
See https://github.com/kbase/relation_engine/runs/5740541694?check_suite_focus=true#step:3:850


ialarmedalien (approved):
LGTM

PR #114, opened 2022-04-29
DATAUP-748 use locale en_US & other cleanup

n1mus:
Change the locale of the `ci::icu_tokenize` analyzer to `en_US` and do some comprehensive-ish tests (a subset of the 10-day tests) to ensure that the index and AQL tokenizer still agree on word boundaries and such.

Also removed the generic/fulltext_search.yaml which I wrote so I know it's not used anywhere because it has performance issues.

For changes to the codebase:


client_src/test/test_integration.py
```diff
@@ -2,7 +2,7 @@
 import os
 from uuid import uuid4
 
-from relation_engine_client import REClient
```
n1mus, line 5: This was confusing since `relation_engine_client` is a directory with an `__init__.py` that defined `__all__` with `main.REClient`

client_src/test/test_integration.py
```diff
@@ -50,6 +50,10 @@ def test_admin_query_invalid_auth(self):
         # Mostly make sure that the __str__ method does not throw any errs
         self.assertTrue("Unauthorized" in str(ctx.exception))
 
+    @unittest.skip(
+        "Insofar inexplicable requests.exceptions.ConnectionError. "
+        "Failure mode for empty query not imperative"
+    )
```
n1mus, line 56: Just ignore this "intermittent" failure

relation_engine_server/utils/json_validation.py
```diff
@@ -159,7 +159,7 @@ def resolve_remote(self, uri):
 
         if scheme in self.handlers:
             result = self.handlers[scheme](uri)
-        elif scheme in [u"http", u"https"]:
+        elif scheme in ["http", "https"]:
```
n1mus, line 162: Autoformatters

spec/stored_queries/generic/fulltext_search.yaml
```diff
@@ -1,94 +0,0 @@
-# Should be REVISED or DEPRECATED.
```
n1mus, line 1: I wrote this and I know it's not being used because of performance issues so axed the cruft

spec/test/stored_queries/test_fulltext_search.py
```diff
@@ -251,148 +248,6 @@ def test_prefix_hit(self):
         )
 
 
-class TestFulltextSearchStoredQuery(unittest.TestCase):
```
n1mus, line 254: Removed tests pertaining to `spec/stored_queries/generic/fulltext_search.yaml`

ialarmedalien (changes requested):
Can you do a little more investigation into the admin query failure? Better still, remove that change from this PR (the rest is fine) and then do some investigation in another PR.

.dockerignore
```diff
@@ -48,4 +48,8 @@ dmypy.json
 
 # docker bits
 Dockerfile*
-docker-compose*
\ No newline at end of file
+docker-compose*
+
+# Temp files
+tmp/
```
n1mus, line 54: These are generated by the comprehensive query testing, or any output files can go here

Makefile
```diff
@@ -13,8 +13,7 @@ shell:
 	docker-compose run re_api sh
 
 reset:
-	docker-compose --rmi all -v
-	docker-compose build
```
n1mus, line 17: Forgot "down", and the "build" is redundant since the other make targets do it

n1mus:
Yep, was planning on that

ialarmedalien (approved):
LGTM

PR #121, opened 2022-05-03
Remove unnecessary push events

ialarmedalien:
Remove the unnecessary run of GHA on push events; now they only run on pull

For changes to the codebase:


.github/workflows/run_tests.yaml
```diff
@@ -1,6 +1,13 @@
 name: Relation Engine test and deploy
 on:
-  [push, pull_request]
+  pull_request:
```
bio-boris, line 3: Shouldn't tests also run on main?

PR #152, opened 2022-07-27
Update to python 3.10

MrCreosote:
- [n/a] I updated the README.md docs to reflect this change.

For changes to the codebase:

- [n/a] I have written tests to cover this change.

### DIFF
--- a/relation_engine_server/test/test_json_validation.py
+++ b/relation_engine_server/test/test_json_validation.py
@@ -271,7 +271,7 @@ class TestJsonValidation(unittest.TestCase):
                     "distance": 3,
                 },
                 "file": "invalid_pattern",
-                "err_str": '"what\'s-the-problem with-this-string\?" does not match .*?',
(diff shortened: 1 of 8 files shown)


ialarmedalien (approved):

relation_engine_server/test/test_json_validation.py
```diff
@@ -350,7 +350,7 @@ def test_date_format_validation(self, schema_arg=None, schema_file_arg=None):
 
         # pyyaml-specific issue: dates get automatically parsed into datetime objects (doh!)
         file_path = os_path.join(json_validation_dir, "unquoted_date.yaml")
-        err_str = "datetime.date\(2020, 6, 6\) is not of type 'string'"
+        err_str = r"datetime.date\(2020, 6, 6\) is not of type 'string'"
```
n1mus, line 353: you could also use `re.escape` which is probably more robust than escaping each individual test string

Issue #156, opened 2022-07-27
Bandit flags `urlopen` in the `json_validator.py`


MrCreosote:
After switching the `urlopen` and `urlsplit` imports in `json_validation.py` from `jsonschema.compat` (which was internal-only and is now removed) to `urllib`, `bandit` complains about using `urlopen` without rigorously checking the url schema. The only reason it didn't before is due to the indirect import in `jsonschema.compat`.

For now, I'm just going to mark it as `# nosec` but we should determine which schemas we need and alter the function to just support those.

https://github.com/kbase/relation_engine/blob/bc5c32d2e367aafe8fd33a627a8fef96a553dbfc/relation_engine_server/utils/json_validation.py#L168

PR #157, opened 2022-07-27
Update dependencies

MrCreosote:
... or alternately, sthu dependabot

This was much more of a pain than I expected when I started on it.

Also see https://github.com/kbase/relation_engine/issues/156

- [n/a] I updated the README.md docs to reflect this change.

For changes to the codebase:

- [n/a] I have written tests to cover this change.


dev-requirements.txt
```diff
@@ -1,10 +1,9 @@
-mypy>=0.630
-bandit==1.5.1
+mypy>=0.971
+bandit==1.7.4
+# mccabe 0.7.0 is not compatible with the flake8 4.0.1
 mccabe==0.6.1
 flake8==4.0.1
-grequests==0.3.0
-coverage==5.2.1
-typed-ast>=1.4.0
```
MrCreosote, line 7: Not needed in python 3.8+

dev-requirements.txt
```diff
@@ -1,10 +1,9 @@
...
-black==22.3.0
-pytest==6.2.5
-jinja2==3.0.3
```
MrCreosote, line 10: Already in `requirements.txt`

relation_engine_server/utils/json_validation.py
```diff
@@ -12,12 +12,10 @@
 """
 from jsonschema import validators, Draft7Validator, FormatChecker, RefResolver
 
-from jsonschema.compat import (
```
MrCreosote, line 15: `jsonschema.compat` was an internal-only API and was removed in 4.X (without a mention in the release notes either...)

n1mus: Hmm I tried to do this before and ran into errors

n1mus: Okay you got the functions from somewhere else. Nice.

relation_engine_server/utils/json_validation.py
```diff
@@ -165,7 +163,8 @@ def resolve_remote(self, uri):
             result = requests.get(uri).json()
         else:
             # Otherwise, pass off to urllib and assume utf-8
-            with urlopen(uri) as url:
+            # TODO https://github.com/kbase/relation_engine/issues/156
```
n1mus, line 166: Seems like this is for the remote specs loading that we don't even use? Maybe? If so, YAGNI for now

scripts/run_tests.sh
```diff
@@ -20,6 +20,6 @@ python -m spec.validate
 python -m relation_engine_server.utils.wait_for api
 # run importer/, relation_engine_server/, and spec/ tests
 coverage run --branch -m pytest
-# RE client tests - seemss like these tests were already run in the line above, why run them again?
+# RE client tests - seems like these tests were already run in the line above, why run them again?
```
n1mus, line 23: Yay

n1mus (approved):
LGTM. Nice!

relation_engine_server/utils/json_validation.py
MrCreosote: Can we confirm we don't use remote specs and we're not going to? If so we could just remove this class (and maybe more stuff)

PR #159, opened 2022-07-29
Add species_or_below flag to sciname queries

MrCreosote:
New field added to the NCBI taxa loader, see https://github.com/kbase/relation_engine_importers/pull/55

Also updated jsonschema

For changes to the codebase:


CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.0.22] TBD
+### Changed
+- The NCBI taxa scientfic name lookup queries below were updated to make use of the new
+  NCBI taxa loader `species_or_below` flag and note that the `strain` flag is deprecated.
+  TODO add link
```
MrCreosote, line 12: TODO before merge: add link to file in https://github.com/kbase/relation_engine_importers/pull/52 when that's merged

spec/stored_queries/taxonomy/taxonomy_search_species_strain.yaml
```diff
@@ -53,7 +53,8 @@ query: |
   )
   FOR doc IN FULLTEXT(@@taxon_coll, @sciname_field, search_text__fulltext)
       FILTER @ts ? doc.created <= @ts AND doc.expired >= @ts : true
-      FILTER doc.rank IN ["species", "strain"] OR doc.strain
+      // note that doc.strain is deprecated TODO add link
```
MrCreosote, line 56: TODO before merge: add link to file in https://github.com/kbase/relation_engine_importers/pull/52 when that's merged

spec/stored_queries/taxonomy/taxonomy_search_species_strain_no_sort.yaml
```diff
@@ -53,6 +53,7 @@ query: |
   )
   FOR doc IN FULLTEXT(@@taxon_coll, @sciname_field, search_text__fulltext)
       FILTER @ts ? doc.created <= @ts AND doc.expired >= @ts : true
-      FILTER doc.rank IN ["species", "strain"] OR doc.strain
+      // note that doc.strain is deprecated TODO add link
```
MrCreosote, line 56: TODO before merge: add link to file in https://github.com/kbase/relation_engine_importers/pull/52 when that's merged

.bandit.yaml
```diff
@@ -0,0 +1,2 @@
+assert_used:
+  skips: ['*_test.py', '*test_*.py']
```
n1mus, line 2: :+1:

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
...
+  - taxonomy_search_species_strain
+  - taxonomy_search_species_strain_no_sort
```
n1mus, line 14: :+1:

scripts/run_tests.sh
```diff
@@ -6,10 +6,14 @@ set -e
 (cd /app/relation_engine_server/test/spec_release && \
   tar czvf spec.tar.gz sample_spec_release)
 
+echo "> black"
```
n1mus, line 9: Curious why not `set -x`

spec/stored_queries/taxonomy/taxonomy_search_species_strain.yaml
n1mus: Curious why is this a code comment? Seems odd to document in code the history of changes. Should it go just into CHANGELOG?

spec/test/stored_queries/test_fulltext_search_flags.py
```diff
@@ -0,0 +1,287 @@
...
+def db(conf, spec):
+    # This is set to the "_system" db by the get_config() method defaults, which are used to start
+    # up the RE API in the docker tests...
+    arango = ArangoClient(conf["db_url"])
+    db = arango.db(conf["db_name"])
+
+    yield db
+
+    # clean up for tests in other modules that might be using the same db / collection
+    _clear_collection(db, spec["name"])
```
n1mus, line 52: Curious. I heard it was un-pythonic to return "complex expressions"

spec/test/stored_queries/test_fulltext_search_flags.py
```diff
@@ -0,0 +1,287 @@
...
+
+@fixture(scope="module")
+def re_client(conf):
+    re_c = REClient(conf["re_api_url"], "non_admin_token")
+    yield re_c
+
+
+@fixture()
+def col(db, spec):
+    yield _clear_collection(db, spec["name"])
```
n1mus, line 63: Curious. I heard it was un-pythonic to return "complex expressions"

spec/test/stored_queries/test_fulltext_search_flags.py
```diff
@@ -0,0 +1,287 @@
...
+from relation_engine_client import REClient
+
+_QUERY_SORT = "taxonomy_search_species_strain"
+_QUERY_NO_SORT = "taxonomy_search_species_strain_no_sort"
+
+
+@fixture(scope="module")
+def conf():
+    config = get_config()
+    config["re_api_url"] = os.environ.get("RE_API_URL", "http://localhost:5000")
```
n1mus, line 26: Wait why is this not in the config already ... :googlymem:

spec/test/stored_queries/test_fulltext_search_flags.py
```diff
@@ -0,0 +1,287 @@
...
+
+
+@fixture(scope="module")
+def re_client(conf):
+    re_c = REClient(conf["re_api_url"], "non_admin_token")
+    yield re_c
+
+
+@fixture()
+def col(db, spec):
```
n1mus, line 62: I like `coll` so as not to confuse with `col` as in `col` and `row`

spec/test/stored_queries/test_fulltext_search_flags.py
```diff
@@ -0,0 +1,287 @@
...
+def _check_queries(re_client, col_name, expected_sorted, search_text):
+    res = _run_query(re_client, col_name, _QUERY_NO_SORT, search_text)
+    assert set(_freeze(res)) == set(_freeze(expected_sorted))
+
+    res = _run_query(re_client, col_name, _QUERY_SORT, search_text)
+    assert res == expected_sorted
+
+
+def _freeze(list_o_dicts):
+    return [frozendict(d) for d in list_o_dicts]
```
n1mus, line 273: What does `frozendict` do

scripts/run_tests.sh
MrCreosote: The tests do fail already if one of these fails, it was just sometimes hard to tell which of the 4 different checkers was failing

spec/test/stored_queries/test_fulltext_search_flags.py
MrCreosote: So 
```
ret = _clear_collection(db, spec["name"])
return ret
```
?

Seems like an extra line without any gain in readability to me. I suppose it depends what "complex expressions" means - to me a function call isn't complex...

spec/test/stored_queries/test_fulltext_search_flags.py
MrCreosote: No idea. This is how it's done in other tests so I just copied it

spec/test/stored_queries/test_fulltext_search_flags.py
MrCreosote: Super simple https://github.com/Marco-Sulla/python-frozendict#introduction

I'm just using it here because the list gets hashed elsewhere when it's put in a `set` so if the `dicts` aren't immutable python will complain

CHANGELOG.md
MrCreosote: done

spec/stored_queries/taxonomy/taxonomy_search_species_strain_no_sort.yaml
MrCreosote: done

spec/stored_queries/taxonomy/taxonomy_search_species_strain.yaml
MrCreosote: Because the next person to look at this is going to wonder why the heck there's two seemingly at least semi-redundant flags in the query and it going to want to remove one. Providing a link to the explanation of how the two flags came to be could save them at best a good chunk of time trying to figure out what's going on and at worst making a mistake that breaks the query.

Done, added link

spec/test/stored_queries/test_fulltext_search_flags.py
MrCreosote: Done

spec/test/stored_queries/test_fulltext_search_flags.py
MrCreosote: Oh wait, it's probably not in the RE API config because the API doesn't need to know it's own URL.

spec/test/stored_queries/test_fulltext_search_flags.py
n1mus: I think the argument is that breaking up any complex expression (realpython's example was mathematical) with temporary variables makes it more readable/robust by computing subexpressions and giving them temporary variables.

Returning a function call may not qualify as a "complex expression", and I've certainly done it. But, I would argue it sort of has the same effect if the function name does not make it clear what is being returned

So, like
```
@fixture()
def coll(db, spec):
    db = _clear_collection(db, spec["name"])
    yield db


def _clear collection(db: arango.database.StandardDatabase, coll_name: str) -> arango.databse.StandardDatabase:
    collection = db.collection(coll_name)
    collection.truncate()

    return collection
```

spec/test/stored_queries/test_fulltext_search_flags.py
n1mus: Oh ...

spec/test/stored_queries/test_fulltext_search_flags.py
n1mus: You could also do this without importing frozendict ... I'm never sure when a 3rd-party package is good but you probably have confidence in this one

MrCreosote: > You could also do this without importing frozendict

That be great, but how? There isn't an equivalent to frozendict in the stdlib (which I've always thought is bizarre, but there you go...)

spec/test/stored_queries/test_fulltext_search_flags.py
MrCreosote: > I think the argument is that breaking up any complex expression (realpython's example was mathematical) with temporary variables makes it more readable/robust by computing subexpressions and giving them temporary variables.

Completely agree, and I think this rule applies everywhere, not just in return statements. However, I don't see a way to break up this function call into smaller parts.

> the function name does not make it clear what is being returned

You're right, I've renamed the function to make it more clear what's going on. My name is slightly different than yours to make it clear that the collection might already exist.

spec/stored_queries/taxonomy/taxonomy_search_species_strain.yaml
n1mus: Oh ... I was just thinking put in the file comments at the top that this query is written de facto to interact with the NCBI imported data and that's what explains the rank filtering criteria. We've never even tried this query on other taxonomy imported data, right?

Putting so much documentation in the code itself seems claustrophobic

spec/test/stored_queries/test_fulltext_search_flags.py
n1mus: I was just relaying the original argument I had seen for not returning complex expressions, which of course don't apply precisely to this situation.

spec/test/stored_queries/test_fulltext_search_flags.py
n1mus: I guess if this is just the general purpose solution to hashing dictionaries, this is a great trick I've learned! And totally the appropriate solution.

I was just thinking in this _exact, isolated_ situation you could also deduplicate the list of dicts instead of casting the list to a set, or other more convoluted rabbit hole ideas that could be investigated

spec/stored_queries/taxonomy/taxonomy_search_species_strain.yaml
MrCreosote: > We've never even tried this query on other taxonomy imported data, right?

Not as far as I know

I feel pretty strongly that the documentation/warning for the code should live as close to the code as possible otherwise it's too easy to miss. This is the kind of stuff that code comments are for, IMO

spec/test/stored_queries/test_fulltext_search_flags.py
MrCreosote: yeah, the equivalent of what's going on here outside of the context of a return would be something like

```
tempvar1 = a + b
tempvar2 = c + d
tempvar3 = tempvar1 + tempvar2
finalvar = tempvar3
```

Here `tempvar3` doesn't help the readability of the code at all and is just taking up space

spec/test/stored_queries/test_fulltext_search_flags.py
MrCreosote: The reason I'm casting to a set is because the result is unordered, and so just deduplicating the list would cause the test to fail. The test just needs to check that the result has the correct elements, regardless of order

spec/test/stored_queries/test_fulltext_search_flags.py
ialarmedalien: Just to play devil's advocate, I don't see any point in this kind of thing:

```
def coll(db, spec):
    coll = _get_cleared_collection(db, spec["name"])
    yield coll
```

This is a needless extra variable and IMO it decreases readability because the function is now on two lines when it could easily be expressed in a single line. There's more of an argument if you wanted to add in some debugging stuff whilst during development, e.g.

```
def coll(db, spec):
    coll = _get_cleared_collection(db, spec["name"])
    # during development, you might want to do something like:
    # print("coll = " + coll)
    new_coll = coll + "_new"
    return new_coll
```

Otherwise I would always advocate for getting rid of unnecessary variables.

Flake8 plugin for removing unnecessary variables before return statements: https://pypi.org/project/flake8-return/

MrCreosote: No disagreement here

spec/stored_queries/taxonomy/taxonomy_search_species_strain.yaml
n1mus: Yeah I agree actually

spec/test/stored_queries/test_fulltext_search_flags.py
n1mus: At this point in my journey I neither agree nor disagree. But when doing something like `return some_func()`, I could only recommend that if `some_func()` clearly indicates what is being returned, usually by function name and function signature documentation.

spec/test/stored_queries/test_fulltext_search_flags.py
n1mus: Ah. Didn't realize pytest had that limitation! unittest doesn't care. In that case ... `sorted(res)` with a key being a value of the document?

spec/test/stored_queries/test_fulltext_search_flags.py
MrCreosote: Agreed, and I think it does that now after you pointed that out - or are there more changes needed IYO?

spec/test/stored_queries/test_fulltext_search_flags.py
MrCreosote: I don't think it's pytest related surely? `assertEquals([1, 2], [2, 1])` would fail just as hard as `assert [1, 2] == [2, 1]`

> sorted(res)

Hm, that would work... it just feels wrong, somehow, compared to the set approach, which is closer to the test criteria, which is 'check that the results returned match these expectations, regardless of order`.

I think of think we've spent way too much time over something pretty trivial at this point, so unless you feel strongly about it I'd say we should just leave it as is

n1mus:
Hopefully the new imports won't slow down the query ...

spec/test/stored_queries/test_fulltext_search_flags.py
n1mus: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual

Yeah, sorry, all good

spec/test/stored_queries/test_fulltext_search_flags.py
MrCreosote: hah, I didn't realize I was replying to AJ and not you earlier until just now

spec/test/stored_queries/test_fulltext_search_flags.py
MrCreosote: Hmm, neat, although that name doesn't really describe what it does.

MrCreosote: I assume you can always use unittest assertions in pytest code...

MrCreosote:
> Hopefully the new imports won't slow down the query ...

I'm not sure how they would? They're in test code that won't even get loaded by the server... and even if they were, it'd be a one time cost at start up and a bit more memory to store the code. Maybe I'm not understanding your point?

If this PR is OK now could you mark it as approved please?

n1mus (approved):
RE: The imports slowing things down. I meant, by importing new data into the `ncbi_taxon` collection, the sciname lookup query speed might scale accordingly

MrCreosote:
Oh gotcha. I'm not super worried about it, each load is < 2m nodes and will happen 1/month at most, but we'll see I guess

PR #184, opened 2025-03-21
SECURITY-49: Add GHA workflows

Xiangs18:


Xiangs18:
**It's not 100% the same, but it's close enough.** 

Before:
Screenshot 2025-03-21 at 5 27 49 PM

Now:
Screenshot 2025-03-21 at 5 28 54 PM

.coveragerc
```diff
@@ -0,0 +1,9 @@
+[coverage:report]
+omit =
+    */test_*
+    relation_engine_server/api_versions/*
+    relation_engine_server/exceptions.py
+    relation_engine_server/main.py
+    relation_engine_server/utils/auth.py
+    relation_engine_server/utils/bulk_import.py
+    relation_engine_server/utils/parse_json.py
```
MrCreosote: Why are these omitted?

.github/codeql.yml
MrCreosote, line 1: Standard question about whether the boilerplate workflows are up to date

.github/workflows/run_tests.yaml
```diff
@@ -11,24 +11,30 @@ jobs:
...
+        docker compose run re_api sh scripts/run_tests.sh
+        docker compose down --remove-orphans
+
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v5
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
+        fail_ci_if_error: true
 
   docker_build_and_push:
```
MrCreosote: Is this still necessary given the other workflows?

.coveragerc
Xiangs18: Because those files are not included in the original coverage, if our goal is not to produce the original coverage, we don't need to omit them.

MrCreosote: How are they excluded from the original coverage?

.github/workflows/run_tests.yaml
Xiangs18: I am not following this. Are you asking if the code coverage upload step is necessary?

.github/codeql.yml
Xiangs18: yes

.github/workflows/run_tests.yaml
MrCreosote: No, I'm asking if docker_build_and_push is necessary

.github/codeql.yml
MrCreosote: :+1:

.github/workflows/run_tests.yaml
Xiangs18: Both Docker Hub and GHA can host images. In that regard, the workflow here is kind of redundant. The only concern I have is whether other services are pulling the latest relation engine image from Docker Hub.

MrCreosote: No services should be pulling images at all. Rancher may be pulling the image, but if you delete that chunk of the yaml the image will still be there, so next time we want to update we just switch docker to GHA

MrCreosote: * other than the catalog, and that's not for core services

.coveragerc
Xiangs18: The original coverage was produced when I added `coverage report --omit=*/test_*` in https://github.com/kbase/relation_engine/pull/184/commits/64926da1f1562ce4bc5fe12d56cd42ea4b684f08. I am not sure why they were excluded from the original coverage, even though most of these files have 0% coverage.

MrCreosote: So there's nothing in the configuration that excludes them in the prior setup? Just `coverage` is dropping them for some reason?

Xiangs18: I did not find any coverage configuration in the codebase that excluded them in the prior setup. What are your thoughts on this? Should we include them if there are no particular reasons to omit these files?

MrCreosote: We definitely shouldn't seemingly arbitrarily exclude files just because the prior setup excluded them

.github/workflows/run_tests.yaml
Xiangs18: 👍

.coveragerc
Xiangs18: 👍

scripts/run_tests.sh
```diff
@@ -10,20 +10,18 @@ echo "> black"
...
 # wait for the RE service to be up so integration tests can pass
 python -m relation_engine_server.utils.wait_for api
-# run importer/, relation_engine_server/, and spec/ tests
-coverage run --branch -m pytest -vv
-# RE client tests - seems like these tests were already run in the line above, why run them again?
-PYTHONPATH=client_src python -m pytest client_src/test
-coverage html --omit=*/test_*
+
+# run importer/, relation_engine_server/, spec/, scripts/, and client_src/ tests
+pytest -vv --cov=. --cov-branch --cov-report=term --cov-report=xml
```
MrCreosote, line 27: Do you not need to exclude the test directory?

Xiangs18: Oops, done.

MrCreosote (commented):
LGTM. @bio-boris do you want to review?

bio-boris (approved):
LGTM

